PR #36, opened 2023-08-23
35 rework merge vex so a SBOM is not required

CBeck-96:
Bugs in merge-vex fixed, merge-vex is now is dedicated to merging only vex files. To merge a vex in an sbom, use the regular merge.


CBeck-96:
Since merge vex is deprecated, PR will be closed until the feature is reworked.

PR #38, opened 2023-09-04
add missing file in cli example Update available_commands.md

andife:


docs/available_commands.md
```diff
@@ -165,8 +165,8 @@ Where name, version, hash and timestamp are information of the `metadata` from a
 
 If it is desired to have a different Regex, this can be done via the flag `--filename-pattern`, i.e.:
 
-    cdx-ev validate mybom.json --filename-pattern=".*" # every character allowed
-    cdx-ev validate mybom.json --filename-pattern="(^bom\.json$)" # only bom.json allowed
+    cdx-ev validate bom.json --filename-pattern=".*" # every character allowed
```
italvi, line 168: I would prefer to remain the "mybom.json" to show that this way also a different name for the bom could be used.

italvi:
Hi @andife thanks for your first contribution and nice catch. Just had one comment.

docs/available_commands.md
andife: ok, changed that

italvi: I was just referring to L#168, I was fine with L#169 being changed to "bom.json", so that a valid file name would be provided for the validation 😁

italvi:
@andife I made an update to the git-workflow and merged a security update. Could you please update your branch and include the change on L#169 with "bom.json"?

italvi (approved):

CBeck-96 (approved):

PR #62, opened 2023-11-07
Add Support for CycloneDx 1.5

CBeck-96:


italvi:
@CBeck-96 I only made some adjustments to the original cdx-files. Otherwise lgtm.

PR #66, opened 2023-11-14
Rework identification of internal components in custom schema

CBeck-96:
Rework according to specification.

should be merged after [Add Support for CycloneDX 1.5](https://github.com/Festo-se/cyclonedx-editor-validator/pull/62)


italvi:
@CBeck-96 please resolve the conflicts. The schema files from the [CycloneDX Specification](https://github.com/CycloneDX/specification/tree/1.5/schema) shall be used. They are in the main through #62.

italvi:
@CBeck-96 the required copyright and group for the metadata component (see [here](https://github.com/Festo-se/cyclonedx-editor-validator/blob/6eb6c2edd813751942cc3bebcb9670a36d7abd5c/cdxev/auxiliary/schema/bom-1.3-custom.schema.json#L184)) does not make sense for the identification of the internal component. I mean, why is it wrong to have a license and not copyright? Why am I required to provide a group? This should be only the case if supplier/publisher/author is Festo, which is the change you included in this PR. Please make this adjustment, too.

italvi:
@CBeck-96 I tried it locally, with different combinations I could think of. The only thing I am uncertain about: What if `copyright` already includes Festo? Then `publisher`, `supplier.name` or `author` is not required to include Festo. But maybe this is something for the plausibility check?

CBeck-96:
> @CBeck-96 I tried it locally, with different combinations I could think of. The only thing I am uncertain about: What if `copyright` already includes Festo? Then, currently, `publisher`, `supplier.name` or `author` is not required to include Festo. But maybe this is something for the plausibility check?

The plausibility check would be a possibility, if no hard enforcement is wished. I am more inclined to integrate it into the schema, as I don't see a scenario where a violation of the rule would be necessary, and it enforces coherence.

italvi (changes requested):
I tested your changes locally and I still have the issue that if `group` includes `com.festo.*`, I am required to have Festo as supplier. This was the prior goal to not have this kind of dependency.

PR #82, opened 2023-12-19
75 add flag ignore missing for set function

CBeck-96:


cdxev/set.py
```diff
@@ -16,6 +16,7 @@
 class SetConfig:
     force: bool
     allow_protected: bool
+    ignore_missing: bool
```
italvi, line 19: why not set it to `false` per default? (must make adjustments to the default provided in the tests, then)

cdxev/set.py
```diff
@@ -223,11 +224,12 @@ def run(sbom: dict, updates: t.Sequence[dict[str, t.Any]], cfg: SetConfig) -> No
...
-            _do_update(target, update, ctx)
+            if not cfg.ignore_missing:
+                msg = LogMessage(
+                    "Set not performed",
+                    f'The component "{update["id"]}" was not found and could not be updated.',
+                )
+                raise AppError(log_msg=msg)
+        if "target_list" in locals():
+            for target in target_list:
+                _do_update(target, update, ctx)
```
italvi, line 235: why not move the whole `for` into the `try`? This would result in the same logic and you would not be required to introduce this new `if`

italvi (changes requested):

cdxev/set.py
```diff
@@ -222,12 +223,12 @@ def run(sbom: dict, updates: t.Sequence[dict[str, t.Any]], cfg: SetConfig) -> No
...
-            raise AppError(log_msg=msg)
-
-        for target in target_list:
-            _do_update(target, update, ctx)
+            if not cfg.ignore_missing:
+                msg = LogMessage(
+                    "Set not performed",
+                    f'The component "{update["id"]}" was not found and could not be updated.',
+                )
+                raise AppError(log_msg=msg)
```
italvi, line 234: I think it would be useful to add an info that a component was not updated, even when `ignore-missing` is used. Otherwise it could suggest that the `set` was done. 

```suggestion
                raise AppError(log_msg=msg)
            else:
                logger.info(
                    LogMessage(
                        "Set not performed",
                        f'The component "{update["id"]}" was not found and could not be updated.',
                    )
                )
```

italvi (changes requested):

italvi (changes requested):
Please also add the documentation for this flag within the `available_commands.md`.

PR #86, opened 2023-12-21
54 validate feature rework dependencies as required

### LINKED ISSUE #54: validate-feature: rework dependencies as required
In some cases, e.g. completely self-written products, there are no third-party components. However, we still require the "dependencies"-field in general. Therefore we should rework the schema to only require the field, if the "components"-field is provided.

CBeck-96:
closes #54


italvi (approved):
Please also add a negative test, i.e. components

PR #101, opened 2024-01-11
feat: amend adds supplier if only publisher or author are given

CBeck-96:
The amend function now adds the supplier field to components, if only publisher or author are given.

### DIFF
--- a/cdxev/amend/operations.py
+++ b/cdxev/amend/operations.py
@@ -131,18 +131,12 @@ class InferSupplier(Operation):
     def handle_component(
         self, component: dict, path_to_license_folder: str = ""
     ) -> None:
-        if (
-            ("supplier" in component)
-        ):
+        if "supplier" in component:
             return
-        if (
-            ("publisher" in component)
-        ):
+        if "publisher" in component:
             component["supplier"] = component["publisher"]
             return
-        if (
-            ("author" in component)
-        ):
+        if "author" in component:
             component["supplier"] = component["author"]
             return
 
--- a/cdxev/amend/operations.py
+++ b/cdxev/amend/operations.py
@@ -135,10 +135,10 @@ class InferSupplier(Operation):
         if "supplier" in component:
             return
         if "publisher" in component:
-            component["supplier"] = component["publisher"]
+            component["supplier"] = {"name": component["publisher"]}
             return
         if "author" in component:
-            component["supplier"] = component["author"]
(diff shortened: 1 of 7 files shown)


cdxev/amend/operations.py
```diff
@@ -132,11 +132,13 @@ class InferSupplier(Operation):
...
     ) -> None:
-        if (
-            ("supplier" in component)
-            or ("publisher" in component)
-            or ("author" in component)
-        ):
+        if "supplier" in component:
+            return
+        if "publisher" in component:
+            component["supplier"] = component["publisher"]
```
italvi, line 138: This assignment should only happen if supplier empty, we should not delete/change information (this would be also according to your documentation, where you state that "if a component does *not* have a supplier".)

cdxev/amend/operations.py
```diff
@@ -132,11 +132,13 @@ class InferSupplier(Operation):
...
+            return
+        if "author" in component:
+            component["supplier"] = component["author"]
```
italvi, line 141: This assignment should only happen if supplier empty, we should not delete/change information (this would be also according to your documentation, where you state that "if a component does *not* have a supplier".)

docs/available_commands.md
```diff
@@ -14,6 +14,9 @@ Currently, the command adds or modifies the following pieces of information:
 
 * If the SBOM metadata doesn't specify an *author* from the SBOM, it will be set to `{"name": "automated"}`.
 * The *compositions* array will be overwritten with a new one which specifies a single *incomplete* aggregate. This aggregate contains all components, including the metadata component.
+* If a component does not have an *supplier*, the tool will try to infer the supplier from the fields (in order of precedence):
```
italvi, line 17: "a" supplier, not "an".

tests/test_amend.py
```diff
@@ -162,6 +162,7 @@ def test_supplier_already_present(self) -> None:
     def test_publisher_already_present(self) -> None:
         component = {"author": "x", "publisher": "y"}
         expected = copy.deepcopy(component)
+        expected["supplier"] = "y"
```
italvi, line 165: This should not be the case. The supplier should not change when already existing.

italvi (changes requested):

tests/test_amend.py
CBeck-96: There is no supplier, only publisher and author

PR #102, opened 2024-01-11
refactor: expiration_field _no_longer_mandatory_in_spec_1.5

CBeck-96:
The expiration field is no longer madatory as pertaining to issue 5940264

### DIFF
--- a/tests/test_validate.py
+++ b/tests/test_validate.py
@@ -558,27 +558,6 @@ class TestValidateUseSchema15(unittest.TestCase):
             issues = validate_test(sbom)
             self.assertEqual(search_for_word_issues("licensee", issues), True)
 
-    def test_no_expiration(self) -> None:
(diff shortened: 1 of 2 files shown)


tests/test_validate.py
```diff
@@ -577,7 +577,7 @@ def test_no_expiration(self) -> None:
                 }
             ]
             issues = validate_test(sbom)
-            self.assertEqual(search_for_word_issues("expiration", issues), True)
+            self.assertEqual(issues, ["no issue"])
```
italvi, line 580: What is the purpose of keeping this test and not removing it completely? It's not like we test for not required fields in general.

italvi (changes requested):

italvi (changes requested):
Please remove the "negative" test and don't have a change not related to the PR

PR #103, opened 2024-01-12
chore: adjust documentation and dependabot according to new branching strategy

italvi:


mmarseu (changes requested):

italvi:
After internal discussion we agreed on dropping this, as it only increases the complexity of the project and could prevent users to contribute. Nevertheless, thanks for your valuable comments, @mmarseu!

PR #115, opened 2024-02-02
Introduce pre-commit

mmarseu:
Hi guys. Take this PR as a suggestion. Since I've been working on #108 - after a long absence from this project, during which I've forgotten all the tools we used to use - I've noticed how much I missed pre-commit hooks here which simply run all the required linters and formatters for me before I can commit and push bad code.  
Obviously, any problems are caught by the actions so they don't get merged into `main` but personally I still prefer catching them as early as possible, i.e., before committing.

This PR adds a dev dependency on [pre-commit](https://pre-commit.com/) and a configuration file which runs all the same tools on every commit which are also used in `.github/workflows/tests.yml` - with the notable exception of pytest because IMO individual commits should be allowed to fail tests ... and also because there is no pre-commit hook for it.  
Three additional hooks are included because they come by default (but could be removed if unwanted):

- `trailing-whitespace` trims trailing whitespace from lines
- `end-of-file-fixer` makes sure every file ends on a newline
- `check-yaml` syntax-checks any yaml files

Of course, there are some minor trade-offs involved here, which you should be aware of:

- Hooks don't work automatically. Every developer must run `pre-commit install` once after `poetry install` to make them work.
  I see this as a positive, really. It's an opt-in system. If you don't want to use the hooks, just don't run that command.
- It takes a few seconds on every commit for the hooks to run. The delay isn't bad but its there and could be annoying for people who do micro-commits.

So please try it out and decide for yourselves whether you think this is a benefit or not. To do so, follow these steps:

1. Check out this branch locally
2. `poetry install`
3. `pre-commit install` (make sure your local venv is active, obviously, or prefix it with `poetry run`
4. `pre-commit run --all-files`
   This step is for you to see the effect. It runs all hooks an all files in the repository and is not normally required. In normal use, hooks are executed on staged files, when you do a commit. 
   You will see that `trailing-whitespace` and `end-of-file-fixer` fix some files. That's not surprising as up to now there is no system in place to avoid violations of these two rules.

For a second test:

1. Change a type annotation somewhere in a python file which should make mypy complain.
2. Add some whitespace to the end of a line in another file. 
3. `git add`
4. `git commit -m "test"`
   You'll get messages about the errors found (whitespace gets fixed automatically, type annotation doesn't) and the commit will fail.

Very importantly, please also make some experiments using the same tools you normally use for working with git. Pre-commit works very well in a terminal, but I think it could be better integrated into VS Code, for example.

#### About `dependabot-pre-commit.yml`

There is a small technicality which prompted me to add a new github workflow. Pre-commit hooks are run in their own isolated python environments where our project dependencies aren't installed. For mypy this means, that some type stubs are missing and the [`--install-types` command-line switch shouldn't be used](https://github.com/pre-commit/mirrors-mypy/blob/main/README.md).

`.pre-commit-config.yaml` allows us to explicitly specify dependencies which get added to the mypy environment **but** that introduces the burden to keep the versions of these type stubs synchronized with `pyproject.toml`.  
Because my goal is not to cause you guys more work, I've added this workflow which *hopefully* updates `.pre-commit-config.yaml` whenever dependabot updates any of the type packages.

I've had no way to test this workflow locally, so I can't really vouch for its correctness.

#### About 9ca9f3f2d68c793465fe8f1ca2811abec630fe5c

This commit isn't directly related to pre-commit. I only noticed during testing of the hooks that the mypy configuration section in `pyproject.toml` wasn't correct.

The `exclude` field wasn't a regex as expected by mypy and wasn't required anyways.  
The `files` field was overly inclusive. Since all our code is organized in a neat package, we can replace this with `packages`  which is simpler (there are no glob patterns to consider).
Again, I can't test Github Actions locally, so I hope these changes don't cause any trouble. I guess I'll see as soon as I create this PR :wink:

### DIFF
--- a/.github/workflows/dependabot-pre-commit.yaml
+++ b/.github/workflows/dependabot-pre-commit.yaml
@@ -1,0 +1,39 @@
+name: Dependabot pre-commit updater
+on: pull_request
+
+permissions:
+  contents: write
+
+jobs:
+  dependabot:
+    runs-on: ubuntu-latest
+    if: github.actor == 'dependabot[bot]'
+    steps:
+      - name: Dependabot metadata
+        id: metadata
+        uses: dependabot/fetch-metadata@v1
+        with:
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
+      - uses: actions/checkout@v4
+        with:
+          # Check out HEAD of source branch. Without this option, the merge commit would be checked out.
+          ref: ${{ github.event.pull_request.head.sha }}
+      - name: Update .pre-commit-config.yaml
+        id: update-pre-commit
+        run: |
+          # Reduces the output of the metadata step above to a simpler JSON format:
+          # { "<package-1-name>": "<new-version>", ... }
+          export updates=$(yq --null-input 'env(updates) | .[] as $item ireduce ({}; .[$item | .dependencyName] = ($item | .newVersion))' -o json)
+
+          # Goes through the dependencies of the mypy hook in pre-commit config and if any of them have
+          # been updated in this PR, replace the version number with the new version
+          yq --inplace 'env(updates) as $up | (.repos[].hooks[] | select(.id == "mypy").additional_dependencies[] | select(split(">=") | .[0] as $pkg | $up | has ($pkg))) |= (split(">=") | .[0] as $pkg | $pkg + "==" + $up[$pkg])' .pre-commit-config.yaml
+
+          # Commit the changes
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+          git add .pre-commit-config.yaml
+          git commit -m "Update additional dependencies for mypy pre-commit hook"
+          git push
+        env:
+          updates: ${{ steps.metadata.outputs.updated-dependencies-json }}
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,33 @@
+﻿# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+default_language_version:
+    python: python3.11
+
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v3.2.0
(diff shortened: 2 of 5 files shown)


italvi:
I ran the tests with `poetry run` and it worked within the environment. However, my default workflow is with Git for Windows and without an environment, hence I installed `pre-commit` via `python -m pip install pre-commit` and the hook worked again for me and threw an error when trying to commit. 

But I am not working with VS Code to commit. @CBeck-96 do you commit with VS Code and could you test it?

italvi:
And your additional workflow was skipped. Is the condition for `on` the right one?

mmarseu:
> And your additional workflow was skipped. Is the condition for `on` the right one?

This is normal. It only runs on Dependabot PRs. The assumption is that these type stub packages are only ever updated by Dependabot. Maybe that's wrong?

mmarseu:
> I ran the tests with `poetry run` and it worked within the environment. However, my default workflow is with Git for Windows and without an environment, hence I installed `pre-commit` via `python -m pip install pre-commit` and the hook worked again for me and threw an error when trying to commit.

True, I hadn't considered that problem. I'm mostly working from the terminal, with the venv activated, so it never occurred to me. 

> But I am not working with VS Code to commit. @CBeck-96 do you commit with VS Code and could you test it?

It does work with VS Code, I always use it. It's just not a great user experience, I think. But at least it picks up the venv automatically, so you shouldn't run into the problem you described above.

AFAIK, you work with PyCharm, right? Does it work there?

italvi:
> AFAIK, you work with PyCharm, right? Does it work there?

No, I am also working with VS Code, but prefer to use a command-line interface to perform `git`-commands. 

As I am not familiar with the workflows for dependabot, I would merge your current approach and as soon as a PR happens, we can test it 😄

italvi (approved):

CBeck-96 (approved):

italvi (approved):

PR #117, opened 2024-02-08
fix: Catch error caused by empty-set logic in pattern

CBeck-96:
The supplier has to contain a "name", "url", or "contact".

If the copyright contains  "([Ff][Ee][Ss][Tt][Oo])" the supplier name is required and has to contain "([Ff][Ee][Ss][Tt][Oo])" as well.


italvi (changes requested):
It seems like there is an issue when copyright includes "Festo".

If supplier includes Festo then the error message is clearly with `ERROR: Invalid SBOM (component: test:test@2.9.1) - 'Copyright (c) Test' does not match '([Ff][Ee][Ss][Tt][Oo])'`

but if `copyright` includes Festo you get with a missing `supplier` the error message `ERROR: Invalid SBOM (component: test:test@2.9.1) - 'supplier' is a required property` and you can add any supplier and it works.

PR #143, opened 2024-03-18
feat: add support for new tools object format from CycloneDX 1.5

mmarseu:
This PR adds support for transforming SBOMs using CycloneDX 1.5's new `.metadata.tools` object.

If the SBOM
- uses the now-deprecated array style, it will stay an array.
- uses the new object style, it will stay an object.
- contains no `.metadata.tools` field, an array will be created.


CBeck-96 (dismissed):

italvi:
> If the SBOM
> * contains no `.metadata.tools` field, an array will be created.

Is this really according to our discussion in #142? Would not be the proper check `if sbom.get("specVersion","") < 1.5 then list`?

cdxev/auxiliary/output.py
```diff
@@ -99,13 +100,38 @@ def update_timestamp(sbom: dict) -> None:
...
-        "name": pkg.NAME,
-        "vendor": pkg.VENDOR,
-        "version": pkg.VERSION,
-    }
+    spec_version = SpecVersion.parse(sbom.get("specVersion", ""))
+    tools: t.Optional[t.Union[dict, list]] = metadata.get("tools", None)
+
+    # Starting in CycloneDX 1.5 tools should be a dict.
+    if tools is None:
+        if spec_version is not None and spec_version >= CycloneDXVersion.V1_5:
```
italvi, line 108: you are comparing ge, but have not defined it in the class

tests/auxiliary/test_spec_version.py
```diff
@@ -0,0 +1,60 @@
+import unittest
+
```
italvi, line 2: Why is this file in the `auxiliary` directory, instead of the root folder `tests`?

italvi (changes requested):
@mmarseu just some small changes + there seems to be a merge conflict in `output.py`

tests/auxiliary/test_spec_version.py
mmarseu: I dunno 🤷 
Because the module to test is also in `auxiliary`. I take it, the tests shouldn't be there?

cdxev/auxiliary/output.py
mmarseu: That's the beaty of [dataclass](https://docs.python.org/3.11/library/dataclasses.html#dataclasses.dataclass) and [total_ordering()](https://docs.python.org/3.11/library/functools.html#functools.total_ordering). I don't have to implement all of the methods myself.

tests/auxiliary/test_spec_version.py
italvi: `test_output` is also in `tests` and not `tests/auxiliary`. Though I understand your reasoning, I would keep the `auxiliary` folder in `tests` for auxiliaries needed for the tests, e.g. SBOMs.

cdxev/auxiliary/sbomFunctions.py
```diff
@@ -2,10 +2,80 @@
...
+    Enumeration members can be compared directly to `SpecVersion` instances.
+    """
+
+    V1_0 = SpecVersion(1, 0)
+    V1_1 = SpecVersion(1, 1)
+    V1_2 = SpecVersion(1, 2)
+    V1_3 = SpecVersion(1, 3)
+    V1_4 = SpecVersion(1, 4)
+    V1_5 = SpecVersion(1, 5)
+
```
italvi, line 62: As we have waited for too long with this PR and now `1.6` is released, maybe add it as well?

italvi (changes requested):
One little request regarding the supported cdx versions, otherwise seems to work as expected 😉

PR #146, opened 2024-03-27
feat: adjust set command to support version ranges

CBeck-96:
Expand set to allow the processing for version ranges.


mmarseu (changes requested):
I suggest you don't attempt to invent your own version range syntax. There are smarter people than us working on that sort of thing and they came up with existing solutions, which additionally fit nicely into the CycloneDX ecosystem. Primarily, purl's proposed version range specification comes to mind:

- [Proposal](https://github.com/package-url/purl-spec/pull/139)
- [Experimental implementation](https://github.com/nexB/univers/)

It is already being [referenced by CycloneDX](https://cyclonedx.org/docs/1.5/json/#vulnerabilities_items_affects_items_versions_items_range)

This would largely render `version_processing.py` obsolete and therefore cuts down the code in this PR by more than half. Its always great to have less code to maintain!

cdxev/auxiliary/identity.py
```diff
@@ -182,3 +184,113 @@ def create(
...
+        swid = Key.from_swid(update["swid"]) if "swid" in update else None
+
+        if allow_unsafe and "name" in update:
+            coordinates = Key.from_coordinates(
+                name=update["name"],
+                group=update.get("group"),
+                version=update.get("version"),
+            )
+        else:
+            coordinates = None
```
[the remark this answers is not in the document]
CBeck-96: removed.

cdxev/auxiliary/identity.py
```diff
@@ -182,3 +184,113 @@ def create(
...
+
+    def __str__(self) -> str:
+        return str(self._keys[0]) if len(self) > 0 else ""
+
+    def __eq__(self, other: object) -> bool:
+        return (
+            isinstance(other, UpdateIdentity)
+            and any(k in self._keys for k in other._keys)
+            and self._version_range == other._version_range
+        )
```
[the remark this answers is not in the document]
CBeck-96: removed.

cdxev/auxiliary/identity.py
```diff
@@ -182,3 +184,113 @@ def create(
             coordinates = None
 
         return ComponentIdentity(cpe, purl, swid, coordinates)
+
+
+@dataclass(frozen=True)
+class UpdateIdentity(ComponentIdentity):
```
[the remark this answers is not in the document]
CBeck-96: removed and replaced by a coordinates subclass that uses version ranges (intended to also be used in the VEX utility).

cdxev/auxiliary/identity.py
```diff
@@ -182,3 +184,113 @@ def create(
...
+
+        version_range = None
+        version_str = update.get("version", "")
+        if is_version_range(version_str):
+            version_range = VersionRange(version_str)
+
+        return UpdateIdentity(cpe, purl, swid, coordinates, version_range=version_range)
+
+    def __str__(self) -> str:
+        return str(self._keys[0]) if len(self) > 0 else ""
```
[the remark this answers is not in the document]
CBeck-96: deleted.

cdxev/auxiliary/identity.py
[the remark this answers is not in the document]
CBeck-96: Agreed, new approach implemented.

docs/available_commands.md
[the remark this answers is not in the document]
CBeck-96: I reworked the Code according to your recommendations.

I think the new documentation is a bit short and heavily relies on the provided links. On one hand, this should be acceptable since we now use the schema officially endorsed by CycloneDX. On the other hand, if you are not familiar with it, you may have to read quite a bit of third-party documentation. However, I am not sure how useful it would be to add a more extensive explanation of the versioning schema in this documentation, or even where to put it.

docs/available_commands.md
```diff
@@ -171,6 +171,80 @@ So only a message that the component was not found and could not be updated is l
 
     cdx-ev set bom.json --from-file mysetfile.json --ignore-missing
 
+#### set for version ranges
```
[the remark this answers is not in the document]
CBeck-96: I extended the existing chapters, but as mentioned above, feels lacking.

cdxev/set.py
```diff
@@ -220,9 +220,15 @@ def run(sbom: dict, updates: t.Sequence[dict[str, t.Any]], cfg: SetConfig) -> No
...
+        target_list: list[dict] = []
         try:
-            target_list = ctx.component_map[update["id"][0]]
+            update_id = update["id"]
+            if isinstance(update_id._keys[0].key, CoordinatesWithVersionRange):
+                for key in ctx.component_map.keys():
+                    if update_id._keys[0] == key:
+                        target_list += ctx.component_map[key]
+            else:
+                target_list = ctx.component_map[update["id"][0]]
```
mmarseu, line 231: Two points:
- Do not access the private `_keys` attribute of the identity object. As a rule, anything that starts with a single or double underscore is private by convention and shouldn't be accessed from the outside. The class is designed as a collection so you can simply write `update_id[0]` as demonstrated in line 231.
- You're storing a reference in `update_id` but you haven't updated line 231 to make us of that same reference. This is not a functional problem, only a stylistic one. It makes it harder for readers to understand that both parts of the code actually reference the same object.  
  Either use the `update["id"][0]` syntax from ln 231 everywhere throughout this block or store a reference to the key (i.e., `update_key = update["id"][0]`) and apply it to every place where it is accessed, including ln 231.

docs/available_commands.md
```diff
@@ -80,6 +80,10 @@ The *target component* can be identified through any of the identifiable propert
 
 If *coordinates* are used to identify the target, they must match the component fully. In other words, if __only__ *name* is given, it will __only match__ components with that name which do __not__ contain *version* or *group* fields.
 
+In *coordinates* it is also possible to provide a range of versions following the [PURl specification](https://github.com/package-url/purl-spec/blob/version-range-spec/VERSION-RANGE-SPEC.rst) as referenced by [CycloneDX](https://cyclonedx.org/docs/1.6/json/#vulnerabilities_items_affects_items_versions_items_range). Further Information on the supported versioning schemas can be found here [univers documentation](https://pypi.org/project/univers/).
+
+It is also possible to use a wildcard by setting the *version* as "*", in that case the version will be ignored and only *name*and *group* will be considered.
```
mmarseu, line 85: Same comment as in `identity.py`. Why this special case? The *vers* specification already allows for the `*` syntax.

docs/available_commands.md
```diff
@@ -129,7 +133,7 @@ When passing the targets, names and values in a file, the file must conform to t
             "id": {
                 # Could be any one of the identifying properties in CycloneDX.
                 # Multiple identifiers are not allowed (with the special exception of name,
-                # group and version which are only valid together)
+                # group and version or version range which are only valid together)
```
mmarseu, line 136: This makes it sound as if *version range* was a different field than *version* (which would arguably be a good idea but isn't the case).

mmarseu (changes requested):
This is a lot better than the last attempt. Good job!
A few comment remain but mostly minor stuff.

docs/available_commands.md
CBeck-96: Changed since its more practical and the initial considerations against it are no longer valid.

docs/available_commands.md
CBeck-96: Because the vers specification allows * only in accordance with the specified versioning schema. 
Since some people use their own versioning schema it was a consideration for this.

Now changed taking recent discussions and 'safety' considerations into account. 
Since univers is open source, people can try to add their schema or switch to a more common one.

cdxev/set.py
CBeck-96: Thanks, adopted `update_key` for all occurences.

cdxev/auxiliary/identity.py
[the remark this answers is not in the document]
CBeck-96: Moved all changes into set.
Now containing two new classes:
UpdateIdentity inherited ComponentIdentity
CoordinatesWithVersionRange inhertited Coordinates

cdxev/auxiliary/identity.py
```diff
@@ -182,3 +195,34 @@ def create(
...
 
         return ComponentIdentity(cpe, purl, swid, coordinates)
+
+
+@dataclass(init=True, frozen=True)
+class CoordinatesWithVersionRange(Coordinates):
+    version_range: univers.version_range.VersionRange = field(
+        default_factory=univers.version_range
+    )
+    version_type: univers.versions = field(default_factory=univers.versions)
```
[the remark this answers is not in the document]
CBeck-96: Thanks, i overlooked that this variable is redundant in the current implementation.

cdxev/auxiliary/identity.py
```diff
@@ -65,7 +68,17 @@ def from_coordinates(
...
-        coordinates = Coordinates(name, group, version)
+        coordinates: Coordinates
+        if version is not None and re.fullmatch("vers:.+/.+", version) is not None:
+            version_range = univers.version_range.VersionRange.from_string(version)
+            version_type = version_range.version_class
+            coordinates = CoordinatesWithVersionRange(
+                name, group, version, version_range, version_type
+            )
+        elif version == "*":
+            coordinates = CoordinatesWithVersionRange(name, group, version, None, None)
```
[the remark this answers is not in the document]
CBeck-96: See comment below. Is removed.

cdxev/auxiliary/identity.py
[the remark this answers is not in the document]
CBeck-96: Removed.

mmarseu:
Is it intentional, that a version range cannot be set via the command-line, only using the `--from-file` option?

cdxev/set.py
```diff
@@ -28,6 +31,105 @@ class Context:
...
+        if isinstance(other, Coordinates):
+            for field_ in fields(other):
+                if field_.name != "version" and getattr(self, field_.name) != getattr(
+                    other, field_.name
+                ):
+                    return False
+
+            if other.version is not None:
+                if (
+                    self.version_range.version_class(other.version)
```
mmarseu, line 67: This is a minor user-experience problem. Assume that the SBOM contains inconsistent data or the user makes a mistake in specifying the scheme of their version range.
This line creates a version object for each component which has the same `name` and `group` as the component we want to update. If the component's version specified in the SBOM cannot be parsed into the versioning scheme the user gave in their range, the tool will crash with an ugly error. 
You should catch that case and either raise an `AppError` (terminates the program with a nice error message) or simply output a warning.

docs/available_commands.md
```diff
@@ -80,6 +80,12 @@ The *target component* can be identified through any of the identifiable propert
 
 If *coordinates* are used to identify the target, they must match the component fully. In other words, if __only__ *name* is given, it will __only match__ components with that name which do __not__ contain *version* or *group* fields.
 
+In *coordinates* it is also possible to provide a range of versions using *version_range* parameter instead *version* following the [PURl specification](https://github.com/package-url/purl-spec/blob/version-range-spec/VERSION-RANGE-SPEC.rst) as referenced by [CycloneDX](https://cyclonedx.org/docs/1.6/json/#vulnerabilities_items_affects_items_versions_items_range).
```
mmarseu, line 83: Typo: _instead **of**_ and _PUR**L**_

docs/available_commands.md
```diff
@@ -80,6 +80,12 @@ The *target component* can be identified through any of the identifiable propert
...
+
+Further Information on the supported versioning schemas can be found here [univers documentation](https://pypi.org/project/univers/). Note that for every schema it is possible to provide a wildcard like *vers:generic/\** to allow all versions of the given schema.
+
+If *version* and *version_range* are provided, *version_range* will be ignored.
```
mmarseu, line 87: This would clearly be an accidental error in the input data, I can't imagine any viable use case for having both fields. Wouldn't it be better to throw a useful error message here, informing the user that they have made a mistake instead of silently doing something which might not be what the user intended?

mmarseu (commented):
Almost good to go!

cdxev/set.py
CBeck-96: I chose to interpret it as "no match" and put out a warning containing the provided version type and the ones the given version is valid under.

docs/available_commands.md
CBeck-96: Fixed

docs/available_commands.md
CBeck-96: That case is now raising an AppError

CBeck-96:
> Is it intentional, that a version range cannot be set via the command-line, only using the `--from-file` option?

It was not...

cdxev/set.py
CBeck-96: But it might be a bit overboard.

mmarseu (dismissed):

docs/available_commands.md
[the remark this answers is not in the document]
mmarseu: I'm not so sure. Looks okay to me. You provide the core bits of info:

- You can use the `version_range` field instead of `version`.
- Ranges must conform to the *vers* specification to be found at your link.

And there are examples, too. I consider this resolved.

docs/available_commands.md
```diff
@@ -129,7 +133,7 @@ When passing the targets, names and values in a file, the file must conform to t
             "id": {
                 # Could be any one of the identifying properties in CycloneDX.
                 # Multiple identifiers are not allowed (with the special exception of name,
-                # group and version which are only valid together)
+                # group and version or version_range which are only valid together)
```
italvi, line 136: I would add a "/" instead of "or", i.e. "version/version_range"

docs/available_commands.md
```diff
@@ -146,7 +150,32 @@ When passing the targets, names and values in a file, the file must conform to t
                         }
                     }
                 ]
-            }
+            },
+        },
+    ]
+
+Example for the use of version ranges:
```
italvi, line 157: Please also add the representation of this component in the SBOM to this example, so people can understand when version_range applies, you can also add one "negative" and "positive" example,  so it is clear when the version range would apply, respectively not.

italvi (changes requested):
@CBeck-96 please change the flag according to our [guidelines](https://github.com/Festo-se/cyclonedx-editor-validator/blob/main/docs/CONTRIBUTING.md#write-new-commands-options-or-arguments) and adjust your example, so that everybody understands when the version_range would apply or not, as otherwise everybody would be required to read the univers-specification.

pyproject.toml
```diff
@@ -26,6 +26,7 @@ python-dateutil = "2.9.0.post0"
 jsonschema = {version = "4.22.0", extras = ["format"]}
 docstring-parser = "^0.16"
 charset-normalizer = "^3.3.2"
+univers = "^30.11.0"
```
italvi, line 29: maybe use a fixed version of univers, just in case the spec is changed? Dependabot and our tests can take care of keeping it up to date and make sure that there are no breaking changes.

CBeck-96: fixed on the current

cdxev/set.py
```diff
@@ -30,6 +34,133 @@ class Context:
...
+    This class inherits from cdxev.auxiliary.identity.Coordinates
+    and extends it to be able to handle version ranges.
+    It achieves this by introducing the attribute 'version_range'
+    of type univers.version_range.VersionRange.
+
+    This class's comparator is compatible with Coordinates objects.
+    If the 'name' and 'group' match,
+    it will check if the provided version of the coordinates object
+    is contained in this class instance's 'version_range'
+    and return True or False depending on the result."
```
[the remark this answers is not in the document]
CBeck-96: removed

cdxev/__main__.py
```diff
@@ -618,6 +618,11 @@ def create_set_parser(
         metavar="<version>",
         help="Version of target component. If specified, name must also be specified.",
     )
+    identifiers.add_argument(
+        "--version_range",
```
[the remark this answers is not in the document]
CBeck-96: changed

docs/available_commands.md
CBeck-96: added two minimal examples

docs/available_commands.md
```diff
@@ -211,6 +211,41 @@ The *target component* can be identified through any of the identifiable propert
...
+
+    vers:generic/>2.0.0
+
+to target all versions ``>2.0.0`` and ``<=4.5.0`` the version range to provide would be
+
+    vers:generic/>2.0.0|<=4.5.0
+
+to target all versions ``>2.0.0`` and ``<=4.5.0`` except ``4.1.1`` version range to provide would be
+
+    vers:generic/>2.0.0|<=4.5.0|>4.1.1|<4.1.1
```
mmarseu, line 233: The canonical representation of this would be `vers:generic/>2.0.0|!=4.1.1|<=4.5.0`

docs/available_commands.md
mmarseu: The documentation should be updated accordingly.

docs/available_commands.md
```diff
@@ -279,16 +317,65 @@ When passing the targets, names and values in a file, the file must conform to t
...
 
     # Perform several operations on properties using set-command
     cdx-ev set bom.json --from-file mysetfile.json
 
+If the file contains a component not present in the SBOM, a error is thrown.
+This can be disabled with the `--ignore-missing` command.
+So only a message that the component was not found and could not be updated is logged.
+
+    cdx-ev set bom.json --from-file mysetfile.json --ignore-missing
+
```
mmarseu, line 378: This part has moved in #185. Take care not to reintroduce it here.

mmarseu (changes requested):

docs/available_commands.md
```diff
@@ -80,6 +80,37 @@ The *target component* can be identified through any of the identifiable propert
 
 If *coordinates* are used to identify the target, they must match the component fully. In other words, if __only__ *name* is given, it will __only match__ components with that name which do __not__ contain *version* or *group* fields.
 
+In *coordinates* it is also possible to provide a range of versions using the *version-range* parameter instead of *version* following the [PURL specification](https://github.com/package-url/purl-spec/blob/version-range-spec/VERSION-RANGE-SPEC.rst) as referenced by [CycloneDX](https://cyclonedx.org/docs/1.6/json/#vulnerabilities_items_affects_items_versions_items_range).
+
+The version range is provided in the schema
+
+    vers:<versioning-scheme>/<version-constraint>|<version-constraint>|...
+
+beginning with the ``vers`` identifier. Following this the versioning-scheme is specified, in the case of semantic versioning this would be ``semver`` or ``generic``. Following this a list of constraints divided by an ``|`` can be provided, to specify which versions are in scope.
```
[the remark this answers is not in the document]
CBeck-96: removed

docs/available_commands.md
```diff
@@ -80,6 +80,37 @@ The *target component* can be identified through any of the identifiable propert
 
 If *coordinates* are used to identify the target, they must match the component fully. In other words, if __only__ *name* is given, it will __only match__ components with that name which do __not__ contain *version* or *group* fields.
 
+In *coordinates* it is also possible to provide a range of versions using the *version-range* parameter instead of *version* following the [PURL specification](https://github.com/package-url/purl-spec/blob/version-range-spec/VERSION-RANGE-SPEC.rst) as referenced by [CycloneDX](https://cyclonedx.org/docs/1.6/json/#vulnerabilities_items_affects_items_versions_items_range).
+
+The version range is provided in the schema
```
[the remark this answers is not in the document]
CBeck-96: Changed to format

docs/available_commands.md
```diff
@@ -211,6 +211,41 @@ The *target component* can be identified through any of the identifiable propert
...
+
+    vers:<versioning-scheme>/<version-constraint>|<version-constraint>|...
+
+beginning with the ``vers`` identifier. Following this the versioning-scheme is specified, in the case of semantic versioning this would be ``semver`` or ``generic``. Following this a list of constraints divided by an ``|`` can be provided, to specify which versions are in scope.
+A few examples:
+
+to target all versions ``>2.0.0`` the version range to provide would be
```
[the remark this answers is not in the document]
CBeck-96: Wote it out as (more opr less readable) sentences

docs/available_commands.md
```diff
@@ -300,7 +296,7 @@ When passing the targets, names and values in a file, the file must conform to t
             "id": {
                 # Could be any one of the identifying properties in CycloneDX.
                 # Multiple identifiers are not allowed (with the special exception of name,
-                # group and version/version_range which are only valid together)
+                # group and version which are only valid together)
```
mmarseu, line 299: Have you intentionally reverted this change?

CBeck-96: No, another remnant from the merge conflict.

mmarseu (dismissed):

italvi (approved):

Issue #156, opened 2024-04-10
What is the point of `merge-vex`?


mmarseu:
The documentation isn't clear on this, so I'd like to ask what the _merge-vex_ command is for.

The documentation simply states:
> This command requires two input files, a SBOM and a VEX file that shell be merged. The VEX file needs to be compatible with the SBOM.

But what is a "VEX file"? That isn't an established term. Even if you google it, you'll find a VEX document only described in an abstract manner, as a set of requirements but not as a complete data format. At the very least, there are currently two implementations which fulfil these requirements and therefore **could** constitute a VEX file: CycloneDX and CSAF. 

Now, if we wanted to merge a CSAF VEX document into the SBOM that would probably justify its own command. At the moment, the command only works for CycloneDX, though, so why do we need a separate command if we already have _merge_ for merging two CycloneDX files?

The only thing the current implementation apparently does differently from the regular _merge_ is check whether all vulnerabilities reference components contained in the SBOM (see also #155). If they don't the command fails silently and simply outputs the original SBOM.  

Could you explain the need for a separate command for this in the documentation?

italvi:
@mmarseu Thanks for starting this discussion as this is something we also discussed internally since #35 and #36. And we decided to deprecate `merge-vex` in favor of a `vex` command. The `vex` command should have several sub-commands.

The ones we currently think of:

- `merge`. This would merge two VEX-files, i.e. not requiring an SBOM anymore. Part of this merge would be to check, whether the same CVE from one VEX-file was maybe updated and therefore the current `analyis.state` should be overwritten (one of the use-cases I can think of there).
- `list`. This would list me all CVEs in the VEX. Here I could even think of further flags, e.g. "--affected" to only see CVEs where (after analysis) I am really affected.
- `prune`/`trim`. This would remove not required entries, e.g. if you want to remove the `false-positives` CVEs from the VEX-file.

I think this command would bring a real benefit and justify a separate command for VEX-files.

mmarseu:
> @mmarseu Thanks for starting this discussion as this is something we also discussed internally since #35 and #36. And we decided to deprecate `merge-vex` in favor of a `vex` command. The `vex` command should have several sub-commands.
> 
> The ones we currently think of:
> 
[...]

PR #157, opened 2024-04-11
tests: Add integration tests

mmarseu:
This PR primarily adds integration testing to test the tool's functionality from a "outside" perspective.
These tests invoke the main function directly so they are about as abstract as can be.

The goal is to ensure that the kind of internal redesign as is planned for #7 will not create any regressions. Existing unit tests are not fit for this because they are so tightly coupled to the design that they, too, will have to be reworked, creating the risk of regressions inside the tests.

Tests for the following commands are implemented:

- [x] amend
- [x] build-public
- [x] merge
- [x] set
- [x] validate

Not to be implemented:

- merge-vex, because it is deprecated (see #156)

#### A few minor additional changes

- Disable E704 for flake because it conflicts with black
- Disable bandit on tests because it makes more trouble than it is worth
- Fix an issue in `test_customreports.py` because I couldn't be bothered to create a separate PR for this. See commit message on 05906d176464589999c57f72bb04190931e788de for details. If you insist, we can still separate this out.
- Fix a bug that made the `--version-range` CLI option to the *set* command not work.


cdxev/set.py
```diff
@@ -192,20 +192,22 @@ def _should_overwrite(
                 "Use the --force option to overwrite."
             ),
         )
+    else:  # pragma: no cover
+        if _prompt_for_overwrite(property, component_id):
+            logger.debug(
+                f'Overwriting "{property}" on component "{component_id}" due to user choice.'
+            )
```
italvi, line 199: Is this information _really_ required? I mean, you will already get the debug-information that the property will be set, feels kind of redundant as this happens through the prompt, i.e. I know that I choose "y"

cdxev/set.py
```diff
@@ -349,12 +350,12 @@ def run(sbom: dict, updates: t.Sequence[dict[str, t.Any]], cfg: SetConfig) -> No
 
     try:
         _validate_update_list(updates, ctx)
-    except AppError:
-        msg = LogMessage(
+    except AppError as e:
+        raise AppError(
```
italvi, line 354: should not this change be applied to [L376](https://github.com/Festo-se/cyclonedx-editor-validator/blob/79b8483d693ad5b10b0c360e81ac7cd7d5530312/cdxev/set.py#L376) too?

tests/test_customreports.py
```diff
@@ -15,31 +17,36 @@ class WarningsNgTestCase(unittest.TestCase):
     def setUpClass(cls) -> None:
         formatter = log.LogMessageFormatter()
         cls.logger = logging.getLogger(__name__)
-        cls.expected_file = "bom.json"
-        cls.expected_target = "issues.json"
+        cls.tempdir = TemporaryDirectory()
+        cls.expected_file = os.path.join(cls.tempdir.name, "bom.json")
+        cls.expected_target = os.path.join(cls.tempdir.name, "issues.json")
         warnings_ng_handler = WarningsNgReporter(
```
italvi, line 23: As you changed the name of the handler in `validate.py`, should it be adjusted to `report_handler`?

tests/integration/helper.py
```diff
@@ -0,0 +1,99 @@
...
+    with path.open() as f:
+        sbom = json.load(f)
+    delete_non_reproducible(sbom)
+    return sbom
+
+
+def delete_non_reproducible(sbom: dict):
+    """
+    Deletes any fields from the SBOM that are typically not reproducible between builds.
+    These would mess lead to wrong test results when comparing actual to expected SBOMs.
```
italvi, line 21: This mess would?

italvi (changes requested):
@mmarseu some small things, the most "critical" one is the missing license in the header of the new `*.py`-files 😁

.flake8
```diff
@@ -2,11 +2,13 @@
 max_line_length = 99
 
 ignore =
-	# Bare except
-	E722
 	# Whitespace before : (conflicting with black, see https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html)
 	E203
+	# Multiple statements on one line (def) : (conflicting with black, see https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html)
```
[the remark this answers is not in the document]
mmarseu: This is a tiny change, so I'm okay adding it to this PR.

cdxev/set.py
mmarseu: Can never have enough debug logging 🤷 😛 

Honestly, I'm fine removing it but its not like that hasn't always been there. I only refactored the code very lightly to remove it from coverage stats.

cdxev/set.py
mmarseu: That's not the same situation. In this line, we previously caught an `AppError` but through away all of the helpful information it contained, instead of exposing it to the user.

In L376 we're building an `AppError` from a `KeyError`. That is not to say that we couldn't maybe use information from the `KeyError` to enrich our own `AppError` but that's a different matter we might or might not look into.

tests/test_customreports.py
italvi: Just for the sake of consistency

italvi (approved):

PR #163, opened 2024-04-22
chore: add notice that merge-vex will be removed in a future version

italvi:
After our discussion in #156 we should add a notice that `merge-vex` will be removed in the future.

### DIFF
--- a/cdxev/__main__.py
+++ b/cdxev/__main__.py
@@ -230,7 +230,7 @@ def create_merge_vex_parser(
     parser = subparsers.add_parser(
         "merge-vex",
         help=(
-            "[Deprecated] - DO NOT USE. This will be removed in a future version."
+            "[Deprecated] - This command will be removed in a future version."
             "Note: The `merge-vex` command will be superseded by a new `vex` command."
             "Merges a VEX file into an SBOM."
         ),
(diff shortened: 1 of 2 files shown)


CBeck-96 (dismissed):

cdxev/__main__.py
```diff
@@ -227,7 +227,14 @@ def create_merge_parser(
 def create_merge_vex_parser(
     subparsers: argparse._SubParsersAction,
 ) -> argparse.ArgumentParser:
-    parser = subparsers.add_parser("merge-vex", help="Merges a VEX file into an SBOM.")
+    parser = subparsers.add_parser(
+        "merge-vex",
+        help=(
+            "[Deprecated] - DO NOT USE. This will be removed in a future version."
```
mmarseu, line 233: I'd remove the `DO NOT USE`. First of all, that's implied by _deprecated_ and secondly, we don't have a replacement, yet. 

Telling people not to use the command without providing them an alternative is a bit weird 😉 

Same goes for `available_commands.md`

mmarseu (dismissed):
I've left you a comment about the message. But since its a minor gripe, you still have my approval, no matter whether you implement it or not.

PR #167, opened 2024-04-24
fix: Disable strict filename validation for default schema

mmarseu:
By default, validation against the default schema is now more lenient with filenames. Warnings are generated for any filename that doesn't match either `bom.json` or `*.cdx.json`. 

Custom filename pattern validation is not affected and remains an error.


cdxev/validator/validate.py
```diff
@@ -169,3 +172,21 @@ def validate_sbom(
                 )
             )
         return 1
+
+
+def validate_filename(
```
italvi, line 177: funny to add this function, after removing it in #108, where it was already in the [helper.py](https://github.com/Festo-se/cyclonedx-editor-validator/pull/108/files#diff-01abb36cd923cdb942c9c9fde54f042831bcd78bcb76283a340195adbbe6fafeL56) 😆

cdxev/__main__.py
```diff
@@ -294,12 +294,14 @@ def create_validation_parser(
     parser.add_argument(
         "--filename-pattern",
         help=(
-            "Regex for validation of file name. "
-            "If no Regex is given the default 'name_version_hash_timestamp.cdx.json' "
-            "or 'bom.json' is used, "
-            "where name, version and timestamp are mandatory and taken from metadata. "
-            "Hash is optional as this is not a required information"
+            "Regex for validation of file name. If not specified, a default regex depending on "
```
italvi, line 297: file name or filename? Both is correct, just would like some consistency, you are mostly using "filename", so I would stick to it

tests/test_validate.py
```diff
@@ -1111,3 +1111,48 @@ def test_internal_component_metadata_copyright_festo_no_supplier(self) -> None:
             sbom["specVersion"] = spec_version
             issues = validate_test(sbom)
             self.assertEqual(search_for_word_issues("supplier", issues), True)
+
+
+class TestValidateFilename(unittest.TestCase):
```
italvi, line 1116: Is not there some redundancy to the tests in [`TestValidateInit`](https://github.com/Festo-se/cyclonedx-editor-validator/blob/3b6899609b48c3676ac92e9910e3e91d05471aba/tests/test_validate.py#L66)?

cdxev/__main__.py
italvi, line 297: I think, you could be more precisely here and state that you are sticking to the [naming convention](https://cyclonedx.org/specification/overview/#recognized-file-patterns) of cdx, if the schema-type is `default` and otherwise it is our regex. Or at least adjust this information in the documentation, but as we would like to try to automatically generate the documentation, I would prefer that you adjust the help text to be more verbose.

italvi (changes requested):
@mmarseu Just some small comments, I think the most critical one is the `no-filename` being ambiguous.

cdxev/validator/validate.py
mmarseu: That's called agile 😁 
I'm not taking this a request to change anything, correct?

cdxev/__main__.py
mmarseu: Very good catch, will fix.

Thanks for generally paying close attention to documentation :pray: It's important and many reviewers skip over it.

tests/test_validate.py
mmarseu: You're right. I didn't think to look in there. Do you know why it's named _Init_?

I propose to delete the old tests. They hook into the validation command at a higher level, so they do test a couple of lines of code that aren't covered by the new tests but those should arguably be covered by integration tests in #157. On the other hand the new tests contain significantly more test cases.
Okay?

cdxev/__main__.py
mmarseu: Mh, I have concerns with that.
 
The generated documentation can never replace what we have in `available_commands.md` because stuffing all of that into the command-line help would be madness. 
Actually, there already is a section in the docs on filename validation and your comment made me notice that. I will update it with the changes made in this PR. So, thanks, even if that wasn't what you had in mind :D

cdxev/validator/validate.py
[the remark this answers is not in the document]
mmarseu: For context, here's my reasoning: IMO it doesn't fit with other functions there. This is part of the validation routine. The others strictly perform non-validation-related tasks (mostly reading files).

That said, I don't mind putting it in helpers 👍

tests/test_validate.py
italvi: As long as you mean to remove only the redundant tests, when yes, I'm fine with it. I would really like to keep the tests for the `warnings-ng` and `gitlab-cq-report`, as these are two essential functions for us, so we should at least know that they are working.

italvi (approved):
@mmarseu I'm fine with the changes, still seems odd that the SBOM is completely compliant, even though you get a warning:
image

But I am fine with it, as the "vanilla" CycloneDX only recommends a naming convention.

PR #181, opened 2024-05-15
docs: add description for commit message format

italvi:


mmarseu:
I've frequently struggled myself with these _types_ of changes. Can we be more precise which type corresponds to a _change of a functionality_ that doesn't truly introduce a new feature but isn't a bugfix either? 
I assume it would be `feat`, right?

Other than that, I propose to shorten it a little and move the format to the top, above the definition of _type_  and _subject_. Want me to commit a proposal?

docs/CONTRIBUTING.md
```diff
@@ -56,3 +57,40 @@ The following things are to consider before submitting a pull request.
...
+
+### Commit Message Format
+
+The message format was mainly inspired by the [guidelines](https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#-git-commit-guidelines) of Angular.
+
+Please use the following format:
+
+```format
+<type>: <subject>
+```
```
mmarseu, line 69: One last point: do we specifically want to discourage a commit body beyond the first line?

That's the way I'm reading the format description right now. If we don't we could add a simple `<optional details>` or something like that.

Otherwise, LGTM.

mmarseu (dismissed):

docs/CONTRIBUTING.md
italvi: like [this](https://github.com/Festo-se/cyclonedx-editor-validator/pull/181/commits/11ca1745302dd91e70e0cd12fb64eea6ac76cb07)?

mmarseu (approved):

PR #191, opened 2024-05-27
feat: Make merge <input> argument optional when `--from-folder` is given

mmarseu:
Overhauls the loading of input files in `invoke_merge()`.

- The user no longer needs to provide direct inputs as positional arguments if `--from-folder` is given and the folder contains at least two input files.
- Input files in the folder are now sorted in the same way the operating system sorts files.
- Inputs contained in `--from-folder` that are **also** specified as positional arguments are no longer merged twice.

**This is a breaking change for two reasons:**

- If the input folder is empty, the program used to exit. It now continues with a warning, if the total number of inputs is 2 or higher.
- The console output has changed slightly. Paths loaded from the input folder are now printed as debug log entries. Any scripts which relied on the paths being printed as before must be changed.

Tests for the new behavior will be added to #157.


CBeck-96 (approved):

CBeck-96 (commented):
Should the documentation be updated?
It can be kept as is, since the information is not wrong, but the case of several input files is not covered.
This might be worth mentioning, since using the input function, the merging order can be influenced (input is merged in order before the folder which contents are merged alphabetically)

mmarseu:
> Should the documentation be updated? It can be kept as is, since the information is not wrong, but the case of several input files is not covered. This might be worth mentioning, since using the input function, the merging order can be influenced (input is merged in order before the folder which contents are merged alphabetically)

Valid point. The documentation should be improved.

That made me think: Is there any point to requiring an input argument when `--from-folder` is given? Since merge order is often important, the user has to make sure to name the files in the folder correctly anyways. Then why do we force them to single out the "primary" merge input, if it could simply be the first file in the folder in alphabetical order?

mmarseu:
> > > Should the documentation be updated? It can be kept as is, since the information is not wrong, but the case of several input files is not covered. This might be worth mentioning, since using the input function, the merging order can be influenced (input is merged in order before the folder which contents are merged alphabetically)
> > 
> > 
> > Valid point. The documentation should be improved.
> > That made me think: Is there any point to requiring an input argument when `--from-folder` is given? Since merge order is often important, the user has to make sure to name the files in the folder correctly anyways. Then why do we force them to single out the "primary" merge input, if it could simply be the first file in the folder in alphabetical order?
> 
> Good point. Now that you mention it, it would be more natural to justb (want to) merge a folders contents.

Done. Tests will be added to #157

italvi (approved):

CBeck-96 (approved):

PR #196, opened 2024-06-03
fix: remove bom-refs only for deleted components in build public

### LINKED ISSUE #154: `build-public` messes up compositions
The *build-public* command does *something* surprising to the `.compositions` array. I haven't checked in the code what exactly it does but it definitely deletes some entries that shouldn't be deleted.

For example, use _build-public_ with a dummy schema that doesn't match anything (see #153) on https://github.com/Festo-se/cyclonedx-editor-validator/blob/8bdf8fd2195b68c18a9f76d74146a8f8d14550a3/tests/auxiliary/test_set_sboms/test.cdx.json#L175-L189

The command shouldn't delete anything and therefore the SBOM should simply remain untouched. However, the result is
```json
"compositions": [
        {
            "aggregate": "incomplete",
            "assemblies": [
                "com.company.unit/depA@4.0.2",
                "some-vendor/depB@1.2.3",
                "depC@3.2.1"
            ]
        }
],
```

CBeck-96:
closes #154

### DIFF
--- a/cdxev/auxiliary/sbomFunctions.py
+++ b/cdxev/auxiliary/sbomFunctions.py
@@ -186,31 +186,23 @@ def get_bom_refs_from_dependencies(dependencies: Sequence[dict]) -> list[str]:
     return list_of_bom_refs
 
 
-def get_ref_from_components(
-    list_of_components: Sequence[dict], only_top_level: bool = True
-) -> list[str]:
+def get_ref_from_components(list_of_components: Sequence[dict]) -> list[str]:
     """
     Function that returns a list of bom-refs from a list of components.
-    Per default only the references of top level components will be retrieved,
-    if the flag 'only_top_level' is set to false,
-    the references of nested components will also be retrieved.
+    This also includes nested components.
 
     Input:
     list_of_components: list with dicts of components
-    only_top_level: bool describing if only top level components shall be considered
 
     Output:
     list_of_bom_refs: List of bom-refs from the components in the submitted list
     """
-    if only_top_level:
-        list_of_bom_refs = []
-        for component in list_of_components:
-            bom_ref = component.get("bom-ref", "")
-            list_of_bom_refs.append(bom_ref)
-        return list_of_bom_refs
-    else:
-        components = extract_components(list_of_components)
-        return get_ref_from_components(components)
+    list_of_all_components = extract_components(list_of_components)
+    list_of_bom_refs = []
+    for component in list_of_all_components:
+        bom_ref = component.get("bom-ref", "")
+        list_of_bom_refs.append(bom_ref)
+    return list_of_bom_refs
 
 
 def extract_components(list_of_components: Sequence[dict]) -> Sequence[dict]:


mmarseu (approved):
As requested, I tried the changes without reviewing the code. LGTM.

cdxev/build_public_bom.py
```diff
@@ -158,14 +158,19 @@ def build_public_bom(sbom: dict, path_to_schema: t.Union[Path, None]) -> dict:
...
-    new_compositions = get_ref_from_components(cleared_components)
+    kept_bom_refs = get_ref_from_components(cleared_components, only_top_level=False)
     remove_internal_information_from_properties(
         sbom.get("metadata", {}).get("component", {})
     )
     sbom["components"] = cleared_components
     sbom["dependencies"] = dependencies
-    compositions = [{"aggregate": "incomplete", "assemblies": new_compositions}]
-    sbom["compositions"] = compositions
+    print(kept_bom_refs)
```
italvi, line 167: debug print?

tests/test_sbom_functions.py
italvi, line 1: Is this really necessary to include this now? I would add this into #157 - there is even already a "sbomFunctionsTests.py".

italvi (changes requested):

tests/test_sbom_functions.py
CBeck-96: I havent had time to look at #157 yet. Where should i put the test instead, (falsly )in test_merge like the others?

mmarseu: #157 is about integration tests. This is a unit test, isn't it?

cdxev/build_public_bom.py
CBeck-96: Propably

tests/test_sbom_functions.py
italvi: Ok, let me change my intention, why this new file? What is the purpose of [sbomFunctionTest.py](https://github.com/Festo-se/cyclonedx-editor-validator/blob/154-build-public-messes-up-compositions/tests/auxiliary/sbomFunctionsTests.py)? Seems to be a forgotten file, @mmarseu?

CBeck-96: Files are cnsolidated

mmarseu: @CBeck-96 To avoid confusion you could rename the filer to `helper.py` or some other commonly used name for that sort of thing.

CBeck-96: Done in #215

italvi (changes requested):
After including the compositions from #154 just for double-check into my SBOM, where the `bom-ref`s are not used for any component, I made the observation that you delete the `bom-ref`s.

Looking at your code, you are using `list_of_components` to get your `bom-ref`s and as they are not within the `components` you remove the `bom-ref`s in `composition`. This is definitely a "plausibility check", not something I would expect from `build-public`. Therefore, please change this behavior.

Do you agree, @mmarseu? (Let's not talk about the irony that we just closed the PR #14 for plausibility check 😆)

mmarseu:
> After including the compositions from #154, just for double-check, into my SBOM, where the `bom-ref`s are not used for any component, I made the observation that you delete the `bom-ref`s.
> 
> Looking at your code, you are using `list_of_components` to get your `bom-ref`s and as they are not within the `components` of my SBOM you remove the `bom-ref`s from #154 in `composition`. This is definitely a "plausibility check", not something I would expect from `build-public`. Therefore, please change this behavior.
> 
> Do you agree, @mmarseu? (Let's not talk about the irony that we just closed the PR #14 for plausibility check 😆)

Yes, I agree. `build-public` should strictly only delete bom-refs for deleted components

italvi (approved):

PR #200, opened 2024-06-10
fix: delete nested components in build public command

### LINKED ISSUE #151: `build-public` incorrectly deletes nested components
The _build-public_ command is meant to delete components marked as _internal_. It isn't documented what it does with non-internal components nested inside those internal components.

The tool should probably either:

1. Delete nested components and remove them from the dependency tree the same way as internal components.
   This is likely the more logical choice, as users might expect components bundled inside internal components to also disappear from the SBOM.
3. Leave nested components in the SBOM and move them up to the parent scope.

Instead, here is what actually happens:

- Delete any component marked as *internal*, **including nested components**.
- Remove dependencies on the internal component.
- **Do not** remove dependencies on the nested components, leaving dependencies to components behind, which aren't part of the SBOM anymore.

We should choose one of the options above, implement it and make it explicit in the documentation.

CBeck-96:
Nested components of a deleted component will now also be removed from the SBOM.

closes #151 

depends on #196

### DIFF
--- a/cdxev/auxiliary/sbomFunctions.py
+++ b/cdxev/auxiliary/sbomFunctions.py
@@ -187,8 +187,7 @@ def get_bom_refs_from_dependencies(dependencies: Sequence[dict]) -> list[str]:
 
 
 def get_ref_from_components(
-        list_of_components: Sequence[dict],
-        only_top_level: bool = True
+    list_of_components: Sequence[dict], only_top_level: bool = True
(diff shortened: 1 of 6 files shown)


mmarseu (approved):
As requested, I tested the changes without reviewing the code. LGTM.

mmarseu (approved):
LGTM

italvi (approved):

PR #201, opened 2024-06-10
fix: do not require a schema for build public

### LINKED ISSUE #153: `build-public` shouldn't require a schema
The *build-public* command requires a `schema-path` option to be provided. The schema serves to identify components that should be deleted. 

This should be made optional. There is the realistic use-case that someone would only want to delete properties in the `internal` namespace without deleting any components.

CBeck-96:
Added the option to not provide a schema.
If no schema is provided the internal properties tagged internal are removed.

closes #153


cdxev/__main__.py
```diff
@@ -625,9 +625,10 @@ def create_build_public_bom_parser(
         type=Path,
     )
     parser.add_argument(
-        "schema_path",
+        "--schema-path",
         metavar="<schema path>",
         help="Path to a json schema, defining when a sbom is considered internal",
```
italvi, line 630: Is this really the correct description? Imo the right wording would be "when **the information** in an SBOM is considered internal".

docs/available_commands.md
```diff
@@ -207,7 +207,13 @@ Currently, two formats are supported: The [warnings-ng-plugin](https://github.co
 
 ## build-public
 
-This command creates a reduced version of an SBOM fit for publication. It:
+This command creates a reduced version of an SBOM fit for publication.
+
+If no JSON schema is provided the command:
+
+* removes any *property* (i.e., item in the `properties` array of a component) whose name starts with `internal:` from all components.
```
italvi, line 214: Not  `name` but `namespace`. I would also add the information that this is done according to https://github.com/CycloneDX/cyclonedx-property-taxonomy#registered-top-level-namespaces

cdxev/build_public_bom.py
```diff
@@ -72,17 +72,24 @@ def remove_component_tagged_internal(
...
-        # if not, only the internal information in properties will be removed
-        if validator_for_being_internal.is_valid(component):
-            list_of_removed_component_bom_refs.append(component.get("bom-ref", ""))
-        else:
+
+    if path_to_schema is not None:
+        validator_for_being_internal = create_internal_validator(path_to_schema)
+        for component in components:
+            # if it is a internal component, the whole component will be removed,
+            # if not, only the internal information in properties will be removed
```
italvi, line 82: "only the property within namespace internal" would be the more appropriate description.

italvi (changes requested):

cdxev/build_public_bom.py
CBeck-96: Changed

docs/available_commands.md
CBeck-96: Changed to "namespace starts with `internal:` from all components, according to [cyclonedx-property-taxonomy](https://github.com/CycloneDX/cyclonedx-property-taxonomy#registered-top-level-namespaces)."

cdxev/__main__.py
CBeck-96: Adopted

docs/available_commands.md
```diff
@@ -207,7 +207,13 @@ Currently, two formats are supported: The [warnings-ng-plugin](https://github.co
 
 ## build-public
 
-This command creates a reduced version of an SBOM fit for publication. It:
+This command creates a reduced version of an SBOM fit for publication.
+
+If no JSON schema is provided the command:
+
+* removes any *property* (i.e., item in the `properties` array of a component) whose namespace starts with `internal:` from all components, according to [cyclonedx-property-taxonomy](https://github.com/CycloneDX/cyclonedx-property-taxonomy#registered-top-level-namespaces).
```
italvi, line 214: "according to CycloneDX property namespace and name taxonomy"

cdxev/build_public_bom.py
```diff
@@ -72,17 +72,24 @@ def remove_component_tagged_internal(
...
-        # if not, only the internal information in properties will be removed
-        if validator_for_being_internal.is_valid(component):
-            list_of_removed_component_bom_refs.append(component.get("bom-ref", ""))
-        else:
+
+    if path_to_schema is not None:
+        validator_for_being_internal = create_internal_validator(path_to_schema)
+        for component in components:
+            # if it is a internal component, the whole component will be removed,
+            # if not, only the only the property within namespace internal will be removed
```
italvi, line 82: only the only

italvi (changes requested):

PR #208, opened 2024-06-17
feat: add more known license names to amend

mmarseu:
This PR only adds a few license names encountered in the wild to the mapping for *amend*.

The diff is so large because the file way indented incorrectly before - it mixed 2 and 4 spaces. To facilitate the review, I split the actual change and the reindent into two commits. You can look at them separately.


CBeck-96 (approved):

italvi (approved):

italvi:
@mmarseu thanks for the two commits, made the review a lot more easier 😅

PR #210, opened 2024-06-17
chore: Add GPL-3.0 license header comment

### LINKED ISSUE #194: fix: add copyright information to files
To be compliant with the [guidelines](https://github.com/Festo-se/cyclonedx-editor-validator/blob/main/LICENSE#L629) of the GPL 3.0, we should add the information, as described there, to all of our files.

mmarseu:
Fixes #194

Additionally, the license has been changed from `GPL-3.0-only` to `GPL-3.0-or-later`. This makes it much easier to keep using our code in the distant future when a hypothetical GPL-4.0 comes around and starts being adopted.
It is safe to assume that later versions of the GPL will not change its spirit which is what counts.


italvi:
@mmarseu Why not simply add the SPDX-ID `GPL-3.0-or-later` in the header as described [here](https://spdx.dev/learn/handling-license-info/#how)?

italvi (changes requested):
Thanks for implementing the change, however, you removed the copyright now. Please add it as well, see this [example](https://github.com/zephyrproject-rtos/zephyr/blob/main/zephyr-env.sh) or this [one](https://github.com/torvalds/linux/blob/master/kernel/backtracetest.c). It's typical to add both.

italvi:
> I've added the copyright line but I'm unsure whether it should really have the years. We've already had this discussion in another issue, I think. It's not legally required and it creates a technical burden to maintain.

I've had a discussion with our legal department regarding this topic and the preferred way is to include the current year of modification, e.g. only change the year, when the file itself was changed, otherwise you can keep the year.

mmarseu:
Does that imply there is no range of years but only the year of the last modification?

Example:
`# Copyright 2024 Festo SE & Co. KG` becomes `# Copyright 2025 Festo SE & Co. KG` the first time the file is touched in 2025?

italvi:
It seems like there are really different opinions, according to the [EU](https://europa.eu/youreurope/business/running-business/intellectual-property/copyright/index_en.htm) placing the year the work was created is sufficient, as you have copyright per default. But looking at best practices from e.g. [GNU](https://www.gnu.org/prep/maintain/html_node/Copyright-Notices.html) they agree on your ranges.

However, looking at the [Berne Convention](https://en.wikipedia.org/wiki/Berne_Convention) maybe you were right from the start and the best thing is to not include copyright at all. I think it would be only required if we have many different contributors from outside of Festo, which is not the case - up to now.

mmarseu:
> It seems like there are really different opinions, according to the [EU](https://europa.eu/youreurope/business/running-business/intellectual-property/copyright/index_en.htm) placing the year the work was created is sufficient, as you have copyright per default.

Let's say it surprised me that a lawyer would argue to *only* have the year of last modification. That's why I asked if that's what they meant. From all I know about copyright law (which isn't very much) that would be shooting ourselves in the foot.

> But looking at best practices from e.g. [GNU](https://www.gnu.org/prep/maintain/html_node/Copyright-Notices.html) they agree on your ranges.

Small wonder since that's where I copied the statement from 😆 

> However, looking at the [Berne Convention](https://en.wikipedia.org/wiki/Berne_Convention) maybe you were right from the start and the best thing is to not include copyright at all. I think it would be only required if we have many different contributors from outside of Festo, which is not the case - up to now.

Copyright automatically applies, no statement is legally necessary. The statement still serves the purpose to make this fact very evident to any reader. I don't think it's entirely useless. Just the year is really not important, I'd say. We have the repository history to prove at what time we have written which code.

On the other hand, if we wanted to claim copyright on the code and be totally above board, we'd have to either not accept outside contributions to this project at all or set up a CLA process. Because if we accept somebody else's code into our codebase and they haven't signed an agreement to sign over the copyright to Festo, then the copyright claim would be wrong. They would technically hold the copyright to their part of the code.
For now that's not a problem, there has never been an outside contributor.

italvi:
> On the other hand, if we wanted to claim copyright on the code and be totally above board, we'd have to either not accept outside contributions to this project at all or set up a CLA process. Because if we accept somebody else's code into our codebase and they haven't signed an agreement to sign over the copyright to Festo, then the copyright claim would be wrong. They would technically hold the copyright to their part of the code. For now that's not a problem, there has never been an outside contributor.

I will discuss this topic further internally and think about introducing a CLA process.

PR #216, opened 2024-07-01
feat: rework validation schema options

mmarseu:
The mutual exclusivity implemented through argparse's `add_mutually_exclusive_group()` method has a subtle and undocumented bug (or call it an unexpected behavior, if you like) that affects options with default values.
If an option with a default value is part of such a group, the option doesn't count as "present on the command-line" if the user explicitly passed its default value.

So in our case, `--schema-type` and `--schema-path` were meant to be mutually exclusive but `--schema-type` has a default value of `default`. So, this invocation correctly raises an error:

    cdx-ev validate --schema-type custom --schema-path myschema.json bom.json

but this invocation does not:

    cdx-ev validate --schema-type default --schema-path myschema.json bom.json

because for options with default values, argparse can't tell the difference between 
- "the option isn't there", and
- "the option has been specified with its default value".

This PR fixes this minor problem by forgoing argparse's built-in default mechanism and replacing it with a hand-coded fallback to the default value in the `invoke_validate()` function.

It includes a few other minor housekeeping changes:
- Update help text for both related command-line options
- Delete unused schema files
- Add `strict` schema type to documentation
- Make `--schema-path` option a proper Path-typed object


cdxev/__main__.py
```diff
@@ -447,31 +447,22 @@ def create_validation_parser(
     group = parser.add_mutually_exclusive_group()
     group.add_argument(
         "--schema-type",
-        help=(
-            "Decide whether to use the default specification of CycloneDX or a custom schema. "
-            "The version will be derived from the specVersion in the provided SBOM. "
-            "If no version is provided defaults to 1.3."
-        ),
+        help="Use a built-in schema for validation.",
```
italvi, line 450: Why did you remove the information that the version is derived from `specVersion`. Now this information is only within the documentation, which is kind of against your wish that persons will only rely on `--help` and also the generation with sphinx won't work then.

cdxev/validator/validate.py
```diff
@@ -22,16 +22,25 @@ def validate_sbom(
...
+    if (schema_path is not None) == bool(schema_type):
+        raise AssertionError(
+            "Exactly one of schema_path or schema_type must be non-None"
         )
 
+    if input_format == "json":
+        try:
+            spec_version = sbom["specVersion"]
+        except KeyError:
+            logger.info("SBOM doesn't specify CycloneDX version. Assuming version 1.3.")
```
italvi, line 39: Do we really want to have this info? As `specVersion` is required by the "vanilla" specification, too, we would now get the same information twice:
image

Maybe this should be on `debug`-level?

italvi, line 39: And: if a `schema-path` is provided, there is no assumption about the `specVersion`, than it's the version from the provided schema.

cdxev/validator/validate.py
```diff
@@ -22,16 +22,25 @@ def validate_sbom(
...
+    filename_regex: t.Optional[str],
+    schema_path: t.Optional[Path],
 ) -> int:
     errors = []
-    if input_format == "json":
-        sbom_schema, used_schema_path = open_schema(
-            sbom, file, schema_type, schema_path
+    if (schema_path is not None) == bool(schema_type):
+        raise AssertionError(
+            "Exactly one of schema_path or schema_type must be non-None"
```
italvi, line 32: Is this even reachable? Now that you removed the default, when both are not `None`, argparse should catch that `schema-type` and `schema-path` are mutually exclusive + if it is not the case, then a test for these lines is missing

italvi (changes requested):
@mmarseu some observations from my side.

Further: If I provide a `JSON`-file, e.g. an SBOM as `schema-path` the program crashes with:
image

Maybe a sanity check, e.g. with checking if something like `$schema": "http://json-schema.org/draft-07/schema#"` is within the provided schema (without the specific schema, so only json-schema.org), is a good idea here?

cdxev/validator/validate.py
mmarseu: No, it's not supposed to be reachable. That's why it's an `AssertionError` and not a pretty error message for the user. It would be a programming bug on the side of the caller if that error was ever raised.
But I admit that in different places I dealt with similar issues in different ways. For instance, same file, L161 there is simply a comment stating that assertion in natural language, no actual test at runtime.
I'm open to removing the check if you prefer.

cdxev/validator/validate.py
mmarseu: I hadn't even considered the fact that `specVersion` is mandatory.
Well, in any case, if we default to a version if it's not specified, then IMO we should inform the user. Otherwise, they might be confused by validation results that aren't applicable to the CDX version they had in mind. To my mind, the only alternative would be to fail early if the version is missing.

mmarseu: You're right, good catch. Thanks!
Will push a fix momentarily.

italvi: "fail early" seems like the wrong decision as we would require an additional validation sequence. And as the "vanilla" specification requires it too, all the tools are also include the field. So the only scenario where a missing `specVersion` could happen, is in case of a manually created SBOM. Maybe we can catch the error from the validation and modify it, like we already do for e.g. [SPDX-IDs](https://github.com/Festo-se/cyclonedx-editor-validator/blob/0369c2143339be5ee03e146e63738c83e5e92577/cdxev/validator/validate.py#L127)?

mmarseu:
> @mmarseu some observations from my side.
> 
> Further: If I provide a `JSON`-file, e.g. an SBOM as `schema-path` the program crashes with: image
> 
> Maybe a sanity check, e.g. with checking if something like `$schema": "http://json-schema.org/draft-07/schema#"` is within the provided schema (without the specific schema, so only json-schema.org), is a good idea here?

This must be what you meant when you said you found an error that has always been there. I was dumb enough to promise I'd fix it 🙄 
Anyways, it's done. User-provided schema are now validated using the proper jsonschema function.

For some reason I started getting a whole lot of mypy errors in `validate.py` unrelated to my changes so I had to fix those, too.
There is a single new mypy ignore comment which is due to an error in the `types-jsonschema` package. If I can get https://github.com/python/typeshed/pull/12484 merged, we can remove it.

cdxev/validator/validate.py
mmarseu: > "fail early" seems like the wrong decision as we would require an additional validation sequence. 

You mean the user would have to fix the error (i.e., add `specVersion`) and then rerun the validator? That's required anyways, unless they by chance targeted the same version that we fall back to. Otherwise, the results are unreliable.

> And as the "vanilla" specification requires it too, all the tools are also include the field. So the only scenario where a missing `specVersion` could happen, is in case of a manually created SBOM. Maybe we can catch the error from the validation and modify it, like we already do for e.g. [SPDX-IDs](https://github.com/Festo-se/cyclonedx-editor-validator/blob/0369c2143339be5ee03e146e63738c83e5e92577/cdxev/validator/validate.py#L127)?

You want to go to that effort for the 0.x% of use cases where somebody authors an SBOM manually and forgets the `specVersion` field? Be my guest, I'm out 😝

cdxev/validator/validate.py
mmarseu: If it's all the same to you, I'll leave it, if only to safe a minute of work-

docs/available_commands.md
```diff
@@ -386,6 +387,7 @@ For all built-in schemas, the tool attempts to determine the correct CycloneDX v
 | Type | Supported CycloneDX versions |
 | ---- | ---------------------------- |
 | `default` | 1.2 to 1.5 |
+| `strict` | 1.2 to 1.3 |
```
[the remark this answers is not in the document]
mmarseu: Oops, I must have gotten something mixed up. My bad. Will restore the files.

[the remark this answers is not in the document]
mmarseu: Done.
FWIW: I'll add integration tests in #157 to catch that problem in future. We should probably have tests for all supported versions.

cdxev/validator/validate.py
italvi: > You mean the user would have to fix the error (i.e., add specVersion) and then rerun the validator? That's required anyways, unless they by chance targeted the same version that we fall back to. Otherwise, the results are unreliable.

Hmm, it does make sense. We even had once the use-case that colleagues manually created SBOMs and wrote `"specVersion": 1.4` but added fields from `1.5`. And then had to re-run the validation anyway. So: Fine for me with the "fail early", better than having the same message twice.

mmarseu: Done

italvi (approved):
@mmarseu after some testing, I found two "bugs", where I think this one is the most critical one: if the JSON-Schema is e.g. an SBOM without `dependencies`-array or just an empty JSON, i.e. `{}`, the validation is successful as `jsonschema` [fallbacks](https://python-jsonschema.readthedocs.io/en/stable/api/jsonschema/validators/#jsonschema.validators.validator_for) to `Draft7Validator` if no `$schema` is found in the provided schema.

But as this is something for the lib itself, I approve the MR. The only thing I could think of: Add a disclaimer in the documentation to be aware of this behavior. But I won't insist on this one, so I leave the decision to you: Merge or change - I can approve it again 😉

mmarseu:
> @mmarseu after some testing, I found two "bugs", where I think this one is the most critical one: if the JSON-Schema is e.g. an SBOM without `dependencies`-array or just an empty JSON, i.e. `{}`, the validation is successful as `jsonschema` [fallbacks](https://python-jsonschema.readthedocs.io/en/stable/api/jsonschema/validators/#jsonschema.validators.validator_for) to `Draft7Validator` if no `$schema` is found in the provided schema.
> 
> But as this is something for the lib itself, 

That's not even a problem of the library, that's just the way JSON Schema is specified. A completely empty JSON object is a valid schema according to the spec. Also, there are no restrictions on additional properties, you can put anything in a schema (probably to support custom extensions). 
So, basically, almost any JSON object could be a valid schema. The only exception are properties with the same names as something found in the JSON Schema meta schema but which don't validate against it &mdash; for instance a field named `dependencies` which exists in both JSON Schema and CycloneDX.

> I approve the MR. The only thing I could think of: Add a disclaimer in the documentation to be aware of this behavior. But I won't insist on this one, so I leave the decision to you: Merge or change - I can approve it again 😉

MERGE!!!

> Just for documentation purposes: The other bug was that the check for the mutual exclusive usage of `schema-type` and `schema-path` only works if a valid choice for `schema-type` is provided, otherwise an error is thrown that `schema-type` is not one of `(default, custom, strict)`. But this is an "issue" with `argparse`.

argparse is far from perfect. That may be of the areas where it could be slightly improved. On the other hand, I imagine it could be very hard to separate between multiple errors in the CLI where each one is legitimate in their own right and simple follow-up errors where you really only have to fix the first and the rest falls into place.

PR #219, opened 2024-07-04
feat: add support for cdx 1.6

CBeck-96:
Adds the offical 1.6 schema as well as a custom version to the tool.

I removed the entire "licensechoice" section in the custom schema, like in 1.5, since the new implementation leads to misleading error messages.


italvi:
Please also add the [updated](https://github.com/CycloneDX/specification/blob/1.6/schema/spdx.schema.json) SPDX list

mmarseu:
> @CBeck-96 please take a look at #157. We decided to remove the not required files, e.g. the `xml`-schemata. And also add integration-tests for `1.6` as done there for `1.2` to `1.5`

Since these required changes were on my head, I went ahead and implemented them. Integration tests are done and the non-JSON-schema files are gone.

tests/test_validate.py
```diff
@@ -525,9 +526,6 @@ def test_use_own_schema_invalid_json(self) -> None:
         )
 
 
-list_of_spec_versions_containing_licensing = ["1.5"]
-
-
 class TestValidateUseSchema15(unittest.TestCase):
```
italvi, line 529: @CBeck-96 you have this specific class for `1.5`, but I think they also apply for `1.6` as you are just testing `licensing` which was introduced with `1.5` and still is valid for `1.6`?

italvi (changes requested):
@CBeck-96 Please answer my question regarding the specific class for tests of `1.5` and rebase.

tests/test_validate.py
CBeck-96: Renamed the Class to reflect its purpose and removed the test that did not align with it

PR #220, opened 2024-07-04
tests: add type hints for validate tests

CBeck-96:
Depends on #219


italvi (changes requested):
@CBeck-96 please rebase, otherwise works for me, if this is what your IDE wants.

italvi (approved):
lgtm

Issue #209, opened 2024-06-17
Custom validation applies components requirements to tools


mmarseu:
This is a side-effect of the change to the `tools` field in CDX 1.5.
When `tools` is an object, it can contain two arrays: `components` and `services`. The official schema applies the regular schemas for those two types to the arrays, which makes sense for them.

In our custom schema that means that all requirements meant for components automatically also apply to tools - for example, `bom-ref`, `copyright`, `license`, etc. This is arguably nonsense and should be changed.

italvi:
Would expect that something like that is covered by our tests 😅

CBeck-96:
The tests only cover fields contained in our CB specification, tools is not among them.

PR #221, opened 2024-07-04
fix: do not apply custom validation components requirements to metadata.tools

CBeck-96:
Depends on #219 and #220


cdxev/auxiliary/schema/bom-1.5-custom.schema.json
```diff
@@ -246,7 +246,7 @@
               "properties": {
                 "components": {
                   "type": "array",
-                  "items": {"$ref": "#/definitions/component"},
+                  "items": {"$ref": "#/definitions/default_component"},
```
italvi, line 249: I would prefer a name like `tools_component`, because it is used within this array, default sounds odd.

italvi (changes requested):
@CBeck-96 only a minor change and please take a look at our tests for `1.5` and `1.6` and adjust them, so they make more sense. Easiest way would be to use the default valid SBOM from @mmarseu in our [integration tests](https://github.com/Festo-se/cyclonedx-editor-validator/blob/main/tests/integration/data/validate/valid/default/laravel_1.5.cdx.json) so you make sure to meet his expectations as described in #209.

italvi (approved):
LGTM, tested it with two SBOMs from colleagues who had issues, which are fixed now.

PR #241, opened 2024-07-23
chore: introduce ossf scorecard

italvi:
Introduce https://github.com/ossf/scorecard-action#installation


mmarseu (approved):
If I understand correctly, this PR only includes a file which has been automatically generated by GitHub? In that case, I see no need to go over the details.
Approved!

PR #247, opened 2024-07-29
fix: remove pinned dependencies according to code scanning

italvi:


mmarseu (approved):
If you find this a useful change, I've got nothing against it 😃

italvi:
@mmarseu, @CBeck-96 please review and approve.

This would fix most issues of https://github.com/Festo-se/cyclonedx-editor-validator/security/code-scanning. However, I have mixed feelings regarding the last two results, which I have not fixed: https://github.com/Festo-se/cyclonedx-editor-validator/security/code-scanning/19 and https://github.com/Festo-se/cyclonedx-editor-validator/security/code-scanning/18. According to https://github.com/ossf/scorecard/issues/4189 we would need a `requirements.txt`

italvi:
@mmarseu I got from one side-quest to another: After moving mkdocs to the `pyproject.toml`, I added tests for the build of gh-pages within the `tests.yaml`. This build threw a warning that one of the anchors was wrong, so I also fix an issue within `CONTRIBUTING.MD" in this PR 😅

mmarseu:
> @mmarseu I got from one side-quest to another: After moving mkdocs to the `pyproject.toml`, I added tests for the build of gh-pages within the `tests.yaml`. This build threw a warning that one of the anchors was wrong, so I also fix an issue within `CONTRIBUTING.MD` in this PR 😅

I know what that's like 😆 

Should we additionally pin the poetry version installed in `main.yml`? Seems like the code scanner thingy doesn't pick up that line because it uses `pipx` instead of `pip` but the problem is exactly the same.

mmarseu (approved):

PR #278, opened 2024-09-07
docs: rework creation of documentation via sphinx

### LINKED ISSUE #177: refactor: change documentation creation from mkdocs to sphinx
IMO, manually adding this content to the docs is bound to lead to outdated docs. If we can get it to auto-generate, I'm all for it 👍 

We could explore sphinx in a separate PR.

_Originally posted by @mmarseu in https://github.com/Festo-se/cyclonedx-editor-validator/pull/160#discussion_r1590940331_

italvi:
Reworked creation of documentation to sphinx, so we can automatically use the help-text
Additionally moved information to readme.
Unfortunately, I could not use the most current version of `sphinx-argparse` as from `0.5.0` it requires at least `Python 3.10`

If you want to build it locally, install sphinx-argparse = "0.4.0", sphinx-rtd-theme = "2.0.0" and sphinx = "7.4.7" (should be installed automatically as dependency of the two extensions) and then execute ` sphinx-build -a -E docs/source/ docs/build/` in the root-directory. `-a` for rewrite and `-E` using new environment. the `build`-folder is excluded by `.gitignore` as this folder should be updated by the `sphinx-build`-command. In the folder you will find all html files.

Unfortunately, I can not add html, so here a screenshot as preview:
image

Closes #177


mmarseu (commented):
Not bad at all!

Have you investigated what can be done about the duplicate labels generated via argparse? As it is, the navigation bar in the HTML output doesn't seem to work correctly because of this.
Also, what of the previous content of `available_commands.md`? There is very valuable information there which isn't part of the command-line help texts.

italvi:
@mmarseu 

> Have you investigated what can be done about the duplicate labels generated via argparse? As it is, the navigation bar in the HTML output doesn't seem to work correctly because of this.

I did and honestly, I am uncertain as according to the [documentation](https://sphinx-argparse.readthedocs.io/en/stable/misc.html#linking-to-action-groups) this should not be an issue. However, looking at the created references, it seems that the created reference is `func-name` rather than `func-subcommand-name`. I could not find anything regarding this in the [project](https://github.com/sphinx-doc/sphinx-argparse).

> Also, what of the previous content of available_commands.md? There is very valuable information there which isn't part of the command-line help texts.

I moved everything to the `README.md`. I referenced in the `README.md` that in available_commands you find more "technical" details. Maybe I should also add a link to the README in the `available_command.rst`?

mmarseu:
> > Have you investigated what can be done about the duplicate labels generated via argparse? As it is, the navigation bar in the HTML output doesn't seem to work correctly because of this.
> 
> I did and honestly, I am uncertain as according to the [documentation](https://sphinx-argparse.readthedocs.io/en/stable/misc.html#linking-to-action-groups) this should not be an issue. However, looking at the created references, it seems that the created reference is `func-name` rather than `func-subcommand-name`. I could not find anything regarding this in the [project](https://github.com/sphinx-doc/sphinx-argparse).

A combination of updating to 0.5.2 and using only a single directive to print the entire help instead of individual directives per sub-command did the trick.
Now, the labels are deduplicated. The warnings are still printed to the console but I think that's a different extension (autolabel or something like that).

As you mentioned, 0.5.0+ no longer supports Python 3.9 but IMO that's not a deal-breaker. We're developing (and generating the docs) on supported Python versions. The tool itself still installs and runs fine on 3.9

> > Also, what of the previous content of available_commands.md? There is very valuable information there which isn't part of the command-line help texts.
> 
> I moved everything to the `README.md`. I referenced in the `README.md` that in available_commands you find more "technical" details. Maybe I should also add a link to the README in the `available_command.rst`?

I had though we might rather merge our docs into the argparse output. It seems the extension already has that feature to add your own content to each section. Do you see a reason not to use it?

README.md
```diff
@@ -4,17 +4,35 @@
...
 [![Imports: isort](https://img.shields.io/badge/%20imports-isort-%231674b1?style=flat&labelColor=ef8336)](https://pycqa.github.io/isort/)
+[![Static Badge](https://img.shields.io/badge/CycloneDX-v1.2%2C1.3%2C1.4%2C1.5%2C1.6-blue?link=https%3A%2F%2Fcyclonedx.org%2Fdocs%2F1.6%2Fjson%2F%23)](https://cyclonedx.org/docs/1.6/json/)
 
 # CycloneDX Editor/Validator
 
-This command-line tool performs various actions on [CycloneDX](https://cyclonedx.org/) SBOMs. It allows you to modify and validate your SBOMs.
+This command-line tool performs various actions on [CycloneDX](https://cyclonedx.org/) SBOMs. It allows you to modify, merge and validate your Software Bill of Materials (SBOM).
 
-## Documentation
+The tool is built with automation in mind, i.e. usage within CI/CD. We try to be a scriptable as possible with various command-line flags to avoid interactive prompts, multiple output options and fine-grained exit codes.
```
CBeck-96, line 13: as scriptable

CBeck-96 (approved):

PR #285, opened 2024-09-19
feat: make validate write to stdout

mmarseu:
The validate command now writes it's output to _stdout_ instead of _stderr_.


italvi (approved):
LGTM, did some checks with also writing stdout from old version to a file and then with the version from this branch and you were right, the file were empty before, now there is content 😉

PR #296, opened 2024-10-11
chore: Add JSON formatting to pre-commit and apply to all

mmarseu:
**This is a pure formatting PR. It does not change any code.**

The reason is that when I tried to edit the CycloneDX schema files, my JSON formatter messed with the entire file and produced a completely unreadable diff, even though the actual change was small.
This PR solves that problem by establishing a base formatting for all JSON files enforced by pre-commit for any future commits.


italvi (approved):
lgtm, changes really only applied to yaml- and json-files.

PR #306, opened 2024-10-14
fix: introduce error handling when version-range is not valid

mmarseu:
Fixes two bugs related to version-ranges.

1. _Set_ used to silently ignore entries with a `version-range` that doesn't parse into _vers_. 
   This PR makes it so in invalid _vers_ string raises an error.
2. Set didn't use to raise an error when a component identified by version-range wasn't found in the SBOM. The component was silently ignored.
   Components targeted by version-range are now treated the same as any other.

Additionally, the `CoordinatesWithVersionRange.version_range` attribute was annotated with the `VersionRange` type but the implementation actually treated it as a `str`. This has also been fixed.


italvi (changes requested):
@mmarseu I could validate your second point, that a missing/wrong `version-range` throws an error but using the cli still throws the error, e.g. `python -m cdxev set bom.json --name "test" --version-range "vers:blub/>=3.0.0" --key "copyright" --value '"blub"'` results in `ValueError: 'vers:blub/>=3.0.0' has an unknown versioning scheme: 'blub'.` Please add this kind of exception-handling as the cli-option should have the same behavior as the `from-file`. We are also missing tests for this case that an invalid `vers` is passed to the `set`-command.

mmarseu:
Thank god _somebody_ is testing this stuff before approving 😆 
I've made this a usage error.

italvi (approved):
I've checked it with different versions via cli and `from-file` option and always received an error, so LGTM.

PR #308, opened 2024-10-16
feat: validate now accepts components with external bom

mmarseu:
The _What_ of this PR is straight-forward: It makes the `validate` command accept components als valid which have an external reference of type `bom`, even if those components do not fulfil the requirements of our custom schema.
The program now outputs a warning to inform the user that the externally referenced BOM cannot be checked.

The _How_ merits a little explanation, I guess, to help with the review. Because the schema diffs look much more confusing than they are.

The custom schemas now wrap all of our custom requirements on components in a big _if-then-else_ construct, where the _if_ tests whether the component has an external reference to a BOM. 
- If it does not, the _else_ clause applies all requirements as before. (see note on `version` below)
- If it does, the _then_ clause applies a schema which is always expected to fail (by requiring an arbitrarily named property). I call this a _marker requirement_.

This _marker requirement_ can be detected in the validation routine. When jsonschema produces an error for this particular marker, the tool only writes a warning but otherwise ignores the error.

**Note on `version`:** If the removal of the `version` requirements trips you up when you see it all the way at the top of the diff, don't worry. I have only moved it into the _else_ clause, because an externally referenced BOM would be where the version is declared.

**Misc changes:**

- Adds a dependency on `pytest-subtests`, a small pytest plugin which enables support for unittest's [subtest feature](https://docs.python.org/3.11/library/unittest.html#distinguishing-test-iterations-using-subtests) when running the tests through pytest. The alternative would be to either not use subtests (but I think the feature is very useful in this case to run tests against all spec versions at once) or to rewrite the tests to native pytest which has a comparable feature.


pyproject.toml
```diff
@@ -43,6 +43,7 @@ types-python-dateutil = "2.9.0.20241003"
 types-jsonschema = "4.23.0.20240813"
 pytest = "8.3.3"
 coverage = "7.6.7"
+pytest-subtests = "^0.13.1"
```
italvi, line 46: I would prefer to use a fixed version and let dependabot do the job. Otherwise we could get a conflict in the lock-file in some PRs, if the version of this package change.

cdxev/validator/validate.py
```diff
@@ -100,7 +100,24 @@ def validate_sbom(
         )
         for error in sorted(v.iter_errors(sbom), key=str):
             try:
-                if len(error.absolute_path) > 3:
+                if error.validator == "required" and error.validator_value == [
+                    "this_is_an_externally_described_component"
```
italvi, line 104: From an implementation point of view: IMO it would be easier to just validate the component and then check here if the component includes an external reference pointing to a SBOM, no? Having this workaround/check in this one file here, would reduce the risk of doing wrong changes in the schema.

italvi (changes requested):
@mmarseu I faced two issues while testing: If e.g. `version` is empty, an error for the component is still thrown, because the `if` is [after](https://github.com/Festo-se/cyclonedx-editor-validator/blob/295-ignore-any-requirements-in-the-custom-schema-on-components-which-have-an-external-reference-of-type-bom-however-output-a-warning-that-the-validity-of-the-component-cannot-be-checked/cdxev/validator/validate.py#L154) the `try-catch`, where you throw the warning, contradicting the statement from the warning.

Further: if the `url` is not a valid `bom-link` as described on the [cdx website](https://cyclonedx.org/capabilities/bomlink/#linking-to-external-boms), respectively [specification](https://cyclonedx.org/docs/1.6/json/#tab-pane_externalReferences_items_url_anyOf_i1_anyOf_i0), you ignore it. So I could write some gibberish, even just an empty `url` and still would ignore the validation. Though this hopefully is only an issue if providing an `externalReference` manually.

pyproject.toml
mmarseu: An oversight on my part. Will be fixed.

mmarseu:
> @mmarseu I faced two issues while testing: If e.g. `version` is empty, an error for the component is still thrown, because the `if` is [after](https://github.com/Festo-se/cyclonedx-editor-validator/blob/295-ignore-any-requirements-in-the-custom-schema-on-components-which-have-an-external-reference-of-type-bom-however-output-a-warning-that-the-validity-of-the-component-cannot-be-checked/cdxev/validator/validate.py#L154) the `try-catch`, where you throw the warning, contradicting the statement from the warning.

I don't really see the problem with this particular behavior. If you have an empty `version` property on a component that references an external SBOM, what information is that supposed to convey? I don't see valid use-cases for this, so the error makes sense.

But your argument brought me to a larger issue I hadn't considered. Obviously, just because the user shouldn't be _required_ to include all of the mandatory fields in a component which references an external SBOM. That is the entire point of this PR.
But what if they _chose_ to include some information nevertheless (e.g., a version as in your example)? Either we prohibit this (not my preferred solution) or we ensure that any properties which _are_ present still conform to our requirements (e.g., not be empty).

AFAIK, this applies to:
- name (minLength 1)
- version (minLength 1)
- copyright (minLength 1)
- licenses (minItems 1)

> Further: if the `url` is not a valid `bom-link` as described on the [cdx website](https://cyclonedx.org/capabilities/bomlink/#linking-to-external-boms), respectively [specification](https://cyclonedx.org/docs/1.6/json/#tab-pane_externalReferences_items_url_anyOf_i1_anyOf_i0), you ignore it. So I could write some gibberish, even just an empty `url` and still would ignore the validation. Though this hopefully is only an issue if providing an `externalReference` manually.

That is by design. Bom-linking isn't the only way to reference an external SBOM. As far as I am concerned, a URL is also fine. The stock CDX spec already ensures that the URL is either a valid IRI-reference or a bom-link. Unfortunately, it is actually not easy to find a string that _isn't_ a valid IRI-reference, so it's hard to verify that this works. Try `:foo` as a URL. You'll see that this throws an error.
The empty string is a special case and I agree that it should be invalid. I'll make the change.

cdxev/validator/validate.py
italvi: After internal discussions, you convinced me for not making this kind of implementation change, as we then would need to differentiate between _custom_ and _stock_ requirements, like e.g. name being still mandatory, so it's not as simple as just ignoring _all_ validation errors if a component has a reference to an external bom. It would increase the complexity and also requiring to differentiate between different cdx versions.

italvi (changes requested):
lgtm, just rebase it please, then I will approve it 😁

italvi (approved):
now lgtm

PR #333, opened 2024-12-02
feat: creates an UUID object for the bomref in init-sbom

### LINKED ISSUE #315: Init-SBOM: automatically create sensible bom-ref
Automatically create a sensible bom-ref like "supplier/name@version" if those values are provided.

CBeck-96:
closes #315


CBeck-96:
Amend could not be reused, since its not compatible, with the cdx model.

tests/integration/test_integration.py
```diff
@@ -299,6 +299,11 @@ def test(
         # Verify that command completed successfully
         assert exit_code == Status.OK
 
+        expected["metadata"]["component"]["bom-ref"] = actual["metadata"]["component"][
+            "bom-ref"
+        ]
+        expected["dependencies"][0]["ref"] = actual["metadata"]["component"]["bom-ref"]
+
```
italvi, line 306: As it's odd to set expected = actual and then later test for `expected == actual`, I would kindly ask you to remove this and replace it with removing the ref completely from the `expected` and `actual` and test just additionally that `actual["metadata"]["component"]["bom-ref]"` is a valid uuid, i.e. utilizing `uuid.UUID(dict)`

italvi (changes requested):
Please make this small adjustments to at least the tests, as it does not feel right to set a value and then check if it's equal, only if this just applies for a small part of the dict.

tests/integration/test_integration.py
CBeck-96: Changed

italvi (approved):

PR #339, opened 2024-12-15
refactor: configure mypy to strict mode

### LINKED ISSUE #11: Configure mypy to strict mode
This would simply increase our code quality.

CBeck-96:
closes #11 

Some errors are excluded, since they will be fixed by transitioning to the cdx python lib


italvi (approved):
tried it locally and the commit hook for mypy works, too.

PR #355, opened 2025-01-07
chore: set minimum required python version to 3.10

italvi:


mmarseu (changes requested):
All of your changes are looking good.
I checked the official migration guide and haven't found any deprecations which would impact us.
However, there are numerous [mentions of `python-version: '3.9'`](https://github.com/search?q=repo%3AFesto-se%2Fcyclonedx-editor-validator%20python-version%3A%20%273.9%27&type=code) in our github workflows. I guess we should be building and testing the code on 3.10 now as well, right?

PR #361, opened 2025-01-13
fix: rearrange instead of delete nested components

### LINKED ISSUE #225: Should `build-public` delete nested components?
The `build-public` command deletes components which are valid according a user-supplied JSON schema. It also deletes all components nested under the deleted component even if those don't match the schema.

Up to now I hadn't given this much thought. I believed it to be intentional. But now I'm reconsidering, because from the point of view of the SBOM consumer, this defeats the SBOMs purpose.

I believe the intention behind `build-public`'s ability to delete entire components is to hide structural or architectural details in the SBOM which are not useful to an authorized consumer and might be considered sensitive by the manufacturer. That's all good. In any case, for vulnerability or license management, this isn't going to make a difference.
But non-internal components aggregated under the internal components *are still part of the product* and they are absolutely important for those use cases.

What do you think? Should `build-public` be changed to move nested components to the parent of the deleted component?

Matze08:
closes #225


cdxev/build_public_bom.py
```diff
@@ -67,28 +77,24 @@ def remove_component_tagged_internal(
...
             # if it is a internal component, the whole component will be removed,
             # if not, the property within namespace internal will be removed
             if validator_for_being_internal.is_valid(component):
                 list_of_removed_component_bom_refs.append(component.get("bom-ref", ""))
-                sub_components = extract_components(component.get("components", []))
-                for comp in sub_components:
-                    list_of_removed_component_bom_refs.append(comp.get("bom-ref", ""))
+                sub_components = component.get("components", [])
+                for sub_component in reversed(sub_components):
+                    components.insert(pos + 1, sub_component)
```
mmarseu, line 92: For instance, a comment would be nice here, especially explaining why you need to reverse the list (might seem obvious to you, because you came up with the algorithm but not to me 😉 )

cdxev/build_public_bom.py
```diff
@@ -39,8 +39,18 @@ def remove_internal_information_from_properties(component: dict) -> None:
...
 
 
+def process_component(component: dict[str, Any]) -> t.List[dict]:
+    cleared_components = []
+    remove_internal_information_from_properties(component)
+    cleared_components.append(component)
+    sub_components = extract_components(component.get("components", []))
+    for sub_component in sub_components:
+        process_component(sub_component)
+    return cleared_components
```
mmarseu, line 49: The list you're returning is guaranteed to have exactly one element: `component`. If that was your intention, please make it clearer by returning `[component]`. Though in that case I would wonder why you're returning a list and not a dict.

cdxev/build_public_bom.py
mmarseu, line 1: Generally, I would love to see a few more comments explaining why your code does what it does. As it is, I have a hard time reviewing this because I'm spending half my time simply trying to understand why you're doing things one way and not another.

mmarseu (changes requested):
Generally, not bad, and certainly at least as good as some of the stuff I've submitted to PR. But when in the position of a reviewer, one is often more critical than when developing one's own stuff, so I will ask you to improve one or two things.
Please consider my comments about missing documentation (i.e., code comments) and the unit tests a suggestion. 
However, the integration test data and the use of deprecated features should be fixed before I approve.

cdxev/build_public_bom.py
```diff
@@ -39,9 +39,36 @@ def remove_internal_information_from_properties(component: dict) -> None:
         component["properties"] = new_properties
 
 
+def clear_component(component: dict[str, Any]) -> dict:
```
mmarseu, line 42: Thank you for your previous change. What I hadn't noticed before (my bad!) is that the function probably shoudn't return anything. 
You're taking one `component` as a parameter and editing it in place. There is no reason to return the dict again and from a design-perspective it is misleading. It implies to callers that a new, modified dict is returned.
Compare, for instance, the [documentation of the `list.sort()` method](https://docs.python.org/3.11/library/stdtypes.html#list.sort)
> To remind users that it operates by side effect, it does not return the sorted sequence (use [sorted()](https://docs.python.org/3.11/library/functions.html#sorted) to explicitly request a new sorted list instance

mmarseu (changes requested):
Thank you for your improvements. There are still a few open points from my previous review and some new ones.

cdxev/build_public_bom.py
```diff
@@ -53,69 +53,69 @@ def clear_component(component: dict[str, Any]) -> dict:
...
+    Removes the components marked as internal (components which are valid
+    according to the schema), from a list of components. The nested
+    components will then replace the removed component if present.
+    If a component is not marked as internal, the function is
+    recursivly checking its nested components.
 
     Parameters
     ----------
     components: list[dict]
         A list of components
```
mmarseu, line 79: This description hasn't been updated to the latest function signature.

cdxev/build_public_bom.py
```diff
@@ -39,57 +39,83 @@ def remove_internal_information_from_properties(component: dict) -> None:
...
+        according to the schema
 
     Returns
     -------
     list[str]
         A list with the bom-refs from the removed
         components
-    list[dict]:
-        A list of components without the property
-        "internal"
```
mmarseu, line 68: This description is incorrect. It lists only the first component of the returned tuple.

mmarseu (commented):
You've still ignored some comments from previous reviews.

cdxev/build_public_bom.py
mmarseu: Now you're saying you're returning two things which isn't possible. What you're returning is a tuple[list[str], list[dict].
You could write, for instance: _A tuple where the first element is a list of bom-refs from the removed components and the second element is a list of the new top-level components_.

mmarseu (changes requested):
Almost there. Only the one docstring left to fix, then I'll approve.

tests/test_build_public_bom.py
```diff
@@ -286,6 +310,45 @@ def test_delete_last_nested_components(self) -> None:
...
+
+    def test_not_delete_internal_properties(self) -> None:
+        component = {
+            "name": "test",
+            "properties": [
+                {"name": "stuff:internal", "value": "still there"},
+                {"name": "stuff", "value": "still there"},
+            ],
+        }
+        expected = component.copy()
```
mmarseu, line 336: Careful, this is a subtle bug. [`dict.copy()`](https://docs.python.org/3/library/stdtypes.html#dict.copy) creates shallow copies. Use [`copy.deepcopy()`](https://docs.python.org/3/library/copy.html#copy.deepcopy), instead.

Easy to prove:
```python
original = {"value": "foo", "list": [1]}
copy = original.copy()
assert original == copy # good, copy and original are identical

copy["list"].append(2) # modify the copy
assert original == copy # ouch, should raise AssertionError but doesn't.
```

In consequence, whatever the function under test does to any nested list or dict inside `component` would also be reflected on `expected`. The assertion would always be true.

mmarseu (changes requested):
I would have loved to approve but your tests include a bug. It's easy to fix 😃

mmarseu (dismissed):

tests/test_build_public_bom.py
italvi, line 1: I am missing a test here, where only an internal property is within a component, i.e.
```json
"properties": [
	{
		"name": "internal:component:status",
		"value": "internal"
	}
]
```
because if you apply `build-public` on it, you keep an empty properties, meaning `"properties": []`, which can be completely removed, as it does not add any value.

italvi (changes requested):
Please see review comments. Just some small grammar issues and one little change with the handling of `properties` if it is empty after removal.

italvi (approved):
thanks, lgtm now!

PR #363, opened 2025-01-13
fix: implemented schema changes to licenses field from CDX 1.5 and 1.6

### LINKED ISSUE #362: Custom schema missed license changes from CDX 1.5 and 1.6
Starting from version 1.5 the custom schema has missed some significant updates to the definition of the `licenses` field. 

CycloneDX 1.5 introduced a `licenseChoice` type which we never introduced to the custom schema. 1.6 added additional fields to licenseChoice which we consequently also don't have.

Here's the diff (left vanilla CycloneDX 1.6, right custom 1.6):
image

mmarseu:
This PR implements the same updates to the `licenses` definition from CycloneDX 1.5 and 1.6 in the custom schema.

From what I could tell, the only modification we made as part of our custom requirements was the `minLength` property on the `expression` property. I've made sure to keep that around.

### DIFF
--- a/cdxev/validator/validate.py
+++ b/cdxev/validator/validate.py
@@ -179,7 +179,8 @@ def validate_sbom(  # noqa: C901
                                 + "used license ID "
                                 + leaf.args[0].split()[0]
                                 + " is not a valid SPDX ID. "
-                                "Please use either the field 'name' or provide a valid ID."
+                                "Please use either the field 'name' and 'text' or "
+                                "provide a valid ID."
                             )
                         elif "non-empty" in leaf.message:
                             errors.append(
--- a/cdxev/validator/validate.py
+++ b/cdxev/validator/validate.py
@@ -149,7 +149,7 @@ def validate_sbom(  # noqa: C901
             except AttributeError:
                 error_path = error.json_path + " has the mistake: "
             if error.context is not None and len(error.context) > 0:
-                if error.validator in ("oneOf", "anyOf") and "licenses" in error.json_path:
+                if error.validator == "oneOf" and "licenses" in error.json_path:
                     # When licenseChoice (array oneOf) fails, the actual errors are nested in
                     # the context of the branch that was closest to passing. Walk the context
                     # tree to find the most relevant leaf errors to surface.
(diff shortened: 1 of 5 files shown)


italvi:
@mmarseu After doing some vibe coding, I think I found the solution to your changes, at least the diff does not seem to be that huge anymore:
image

And all tests are passing now. If you could review if I kept your original idea of solving #362 that would be great. Then we would just need @CBeck-96 to approve it, as you and I contributed to this branch.

cdxev/validator/validate.py
```diff
@@ -150,22 +149,70 @@ def validate_sbom(  # noqa: C901
...
+                    all_leaves: list[jsonschema.exceptions.ValidationError] = []
+                    seen_messages: set[str] = set()
+                    for ctx_error in error.context:
+                        for leaf in collect_leaf_errors(ctx_error):
+                            if leaf.message not in seen_messages:
+                                seen_messages.add(leaf.message)
+                                all_leaves.append(leaf)
+
+                    if not all_leaves:
+                        all_leaves = collect_leaf_errors(error.context[0])
```
mmarseu: IMO completely unnecessary. I don't see a code path where this condition could be true.

cdxev/validator/validate.py
```diff
@@ -150,22 +149,70 @@ def validate_sbom(  # noqa: C901
...
+                    for leaf in all_leaves:
+                        if (
+                            "license.id" in leaf.json_path or "license.id" in error.json_path
+                        ) and ("is not one of" in leaf.message):
+                            errors.append(
+                                error_path
+                                + "used license ID "
+                                + leaf.args[0].split()[0]
+                                + " is not a valid SPDX ID. "
+                                "Please use either the field 'name' or provide a valid ID."
```
mmarseu: Small improvement to the error message: "Please use either the fields 'name' and 'text' or ..."

mmarseu (commented):
I can't really verify all of the changes you made. As long as the tests pass, I guess it should be fine.

mmarseu:
I've made a few small changes to the schema files, mostly to reduce the differences between the stock and custom schema. Only 5bfa4a1c2883b0686fd2bbbf6fe1f6933e1936da contains a change which might have an impact on error messages produced by the validator. The tests pass, though.

cdxev/validator/validate.py
italvi: Implemented in https://github.com/Festo-se/cyclonedx-editor-validator/pull/363/changes/b93ebf34ea6a5969497ad632593e24517a112961

cdxev/validator/validate.py
italvi: done in https://github.com/Festo-se/cyclonedx-editor-validator/pull/363/changes/d0b6882a67318546871ed5768c3af23d9bea65f7

italvi (approved):

CBeck-96 (approved):

Issue #224, opened 2024-07-11
Build public does not delete metadata.component even if its tagged internal


CBeck-96:
Add this fact to the documentation.
Maybe catch the use of "internal.component.status = internal" with an error/warning in build public and validate.

PR #364, opened 2025-01-13
feat: throw warning in build-public if schema applies to metadata.component 

Matze08:
closes: #224


cdxev/build_public_bom.py
```diff
@@ -1,14 +1,47 @@
...
 
 from jsonschema import Draft7Validator, FormatChecker
 
 from cdxev.auxiliary.sbomFunctions import extract_components
+from cdxev.log import LogMessage
+
+logger = logging.getLogger(__name__)
+
+
+def check_affected_metadata(metadata: dict[str, Any], path_to_schema: Path) -> bool:
```
italvi, line 18: Please use the already existing function, where it's checked for `components` instead of introducing a new function

cdxev/build_public_bom.py
```diff
@@ -1,14 +1,47 @@
...
+    bool
+        True metadata.component is affected,
+        False if not
+    """
+    validator_for_being_internal = create_internal_validator(path_to_schema)
+    if validator_for_being_internal.is_valid(metadata.get("component", [])):
+        logger.warning(
+            LogMessage(
+                "Warning: `metadata.component` is not affected by the JSON schema!",
+                "Please check manually.",
```
italvi, line 40: The warning should be something like "`metadata.component` was not removed even though the JSON schema applies to it. Maybe you try to create an external SBOM for an internal component?"

italvi (changes requested):

cdxev/build_public_bom.py
```diff
@@ -203,13 +222,21 @@ def build_public_bom(sbom: dict[str, Any], path_to_schema: t.Union[Path, None])
...
+        sbom["components"] = cleared_components
+    else:
+        sbom.pop("components", None)
     for bom_ref in list_of_removed_component_bom_refs:
         dependencies = merge_dependency_for_removed_component(bom_ref, dependencies)
     remove_internal_information_from_properties(
         sbom.get("metadata", {}).get("component", {})
     )
-    sbom["dependencies"] = dependencies
+    # replace dependencies with new dependencies, if it is not  an empy list
```
italvi, line 235: one whitespace too much between "not" and "an".

docs/source/usage/build-public.rst
```diff
@@ -16,6 +16,10 @@ build-public
     The actions are performed in this order, meaning that *internal* properties will be taken into account when matching the JSON schema.
     If a component containing nested components is deleted, those nested components are deleted as well.
 
+    .. warning::
+        **Warning!** The ``metadata.component`` will not be removed even though the JSON schema applies to it. This is because no important information may be deleted.
```
italvi, line 20: The description is wrong. The "even though" does not make sense, should be "even when", as the schema does not have to apply to it. Further, the reasoning should be "This is because the ``metadata.component`` is the component the BOM describes, therefore removing it, would make the SBOM ambiguous."

italvi (changes requested):
Just some small comments regarding grammas and the documentation.

docs/source/usage/build-public.rst
```diff
@@ -16,6 +16,11 @@ build-public
     The actions are performed in this order, meaning that *internal* properties will be taken into account when matching the JSON schema.
     If a component containing nested components is deleted, those nested components are deleted as well.
 
+    .. warning::
+        **Warning!** The ``metadata.component`` will not be removed even when the JSON schema applies to it.
```
italvi, line 20: I would remove the "Warning" as it is sufficient to have the "warning"-box:
image

docs/source/usage/build-public.rst
```diff
@@ -16,6 +16,11 @@ build-public
     The actions are performed in this order, meaning that *internal* properties will be taken into account when matching the JSON schema.
     If a component containing nested components is deleted, those nested components are deleted as well.
 
+    .. warning::
```
italvi, line 19: Please also add "However, this behavior does not affect the deletion of any property, which starts with the name ``internal:``", as otherwise it could be misleading.

Further, please also change the order of the bullet points at the beginning of the file, as the deletion always happen, wheras the schema is optional:
image

docs/source/usage/build-public.rst
```diff
@@ -10,8 +10,8 @@ build-public
 
     This command creates a redacted version of an SBOM fit for publication. It:
 
-    * can optionally delete entire components matching a JSON schema provided by the user, and it
     * deletes any *property* (i.e., item in the ``properties`` array of a component) whose name starts with ``internal:`` from all components.
+    * can optionally delete entire components matching a JSON schema provided by the user, and it
```
italvi, line 14: Now the "and it" at the end of the sentence does not fit, please either write the "and it" at the end of the first bullet point or remote it completely.

docs/source/usage/build-public.rst
```diff
@@ -20,6 +20,7 @@ build-public
         The ``metadata.component`` will not be removed even when the JSON schema applies to it.
         This is because the ``metadata.component`` is the component the BOM describes, therefore removing it, would make the SBOM ambiguous.
         If the schema applies to the ``metadata.component``, the SBOM is likely not intended for public use.
+        However, this behavior does not affect the deletion of any property, which starts with the name ``internal:``
```
italvi, line 23: dot at the end of the sentence

italvi (changes requested):

italvi (approved):
now it looks good, thank you! 😄

PR #342, opened 2024-12-16
feat: add 'vex' option

Matze08:
closes #156

PR #404, opened 2025-04-09
refactor: use generalized input argument for commands

CBeck-96:
To be merged after #342


italvi (changes requested):
- Build public prints "Path to a SBOM file.", should be either the (now) default text or "an SBOM".
- You removed the custom helper text for the input of `validate` but still introduced the one for `merge`. Please either decide to just have one helper for every function or leave custom text as it is.

CBeck-96:
> * Build public prints "Path to a SBOM file.", should be either the (now) default text or "an SBOM".
> * You removed the custom helper text for the input of `validate` but still introduced the one for `merge`. Please either decide to just have one helper for every function or leave custom text as it is.

1. The whole command should have been replaced -> done
2. Merge requires the information that at least two paths have to be provided, while validate only needs one sbom and the additional information does not provide real benefit.

cdxev/__main__.py
```diff
@@ -1159,8 +1117,9 @@ def invoke_vex(args: argparse.Namespace) -> int:
 
     if args.sub_command == "list":
         output = vex(sub_command=args.sub_command, file=file, schema=args.schema)
-        write_list(str(output), args.output, file, format=args.format)
 
+    if args.sub_command == "list":
+        write_list(str(output), args.output, file, format=args.format)
```
italvi, line 1122: You reintroduced the abundant double-if

italvi (changes requested):

cdxev/__main__.py
CBeck-96: Removed

italvi (changes requested):
You have not adjusted the function for the `list`-command, resulting in the following output:
image
in contrast to e.g. `amend`:
image

Notice the square brackets in the usage:, because you provide a custom usage description in the code + you add the input in a separate argument, instead of using the introduced input function

CBeck-96:
> You have not adjusted the function for the `list`-command, resulting in the following output: image in contrast to e.g. `amend`: image
> 
> Notice the square brackets in the usage:, because you provide a custom usage description in the code + you add the input in a separate argument, instead of using the introduced input function

The reason for the [] lies in the optionality of all the inputs for amend (even input), while in list, the input and operation are mandatory arguments

docs/source/CONTRIBUTING.rst
italvi, line 1: This should not be here. GitHub did something strange and just added my PR #421 into this branch. Please rebase this PR with main.

italvi:
Will be resolved now in #426.

PR #411, opened 2025-04-24
fix: correct errors in component comparison

### LINKED ISSUE #409: Identity should use are hierarchy with its keys
The Identity class uses the "any" operation to compare its keys. If two components have the same name and version but different PURLs, they are still considered identical even though they might be from different ecosystems.

This is only possible using the "allow unsafe" option, but in practice, a "safe" operation is not practical. Therefore, the comparison should have a hierarchy.
The function should check if both components contain a CPE and/or a PURL.
If those are only present in one (or none), it should proceed to compare the name and version.

CBeck-96:
closes #409 

Extends the SWID object and makes the comparison hierarchical.


mmarseu:
FYI, this PR (just like my original implementation before 😏) still breaks the [`__hash__()` method's contract](https://docs.python.org/3.12/reference/datamodel.html#object.__hash__): 
> The only required property is that objects which compare equal have the same hash value

This means we should expect trouble when using this class as a dict key.

In a nutshell, you cannot make `__eq__()` compare a dynamically selected subset of an object's properties if you want to make the object hashable (for use in dict keys). 

For example, all of the following three objects would compare equal (i.e., `object.__eq__(other) == True`).
```python
# Object 1
{
  name: "foo",
  version: "1.0",
  purl: "pkg:npm/foo@1.0"
}

# Object 2
{
  purl: "pkg:npm/foo@1.0",
  cpe: "cpe:2.3:a:acme:foo:1.0:*:*:*:*:*:*:*"
}

# Object 3
{
  cpe: "cpe:2.3:a:acme:foo:1.0:*:*:*:*:*:*:*",
  swid: {
    tagId: "acme:foo@1.0",
    name: "foo",
    version: "1.0"
  }
}
```

How would you implement `__hash__()` to create the same hash for all three?

cdxev/auxiliary/identity.py
```diff
@@ -75,17 +83,24 @@ class SWID(dict):
     """
     SWIDs are a complex construct which can contain a lot of information.
 
-    Fortunately, a single property of an SWID is both mandatory and unique: the tagId.
+    This class employs a naive approach to SWID equality. Two SWIDs are considered
+    to refer to the same component iff the required fields tagId and name are identical
```
italvi:
```suggestion
    to refer to the same component if the required fields tagId and name are identical
```

tests/test_identity.py
```diff
@@ -188,6 +188,69 @@ def test_id_equality(self) -> None:
...
+        self.assertNotEqual(id_base, id_cpe)
+        self.assertNotEqual(id_base, id_purl)
+        self.assertNotEqual(id_base, id_swid)
+        self.assertNotEqual(id_base, id_coords)
+
+    def test_hash_contract(self) -> None:
+        # Objects that compare equal must have the same hash value (Python's hash contract).
+        # This is non-trivial here because two equal ComponentIdentities may have different
+        # key sets (e.g. one with only PURL, another with PURL+CPE).
+        purl_only = ComponentIdentity(Key.from_purl(self.sample_purl))
```
italvi, line 233:
```suggestion
    def test_hash_contract(self) -> None:
        purl_only = ComponentIdentity(Key.from_purl(self.sample_purl))
```

tests/test_identity.py
```diff
@@ -188,6 +188,69 @@ def test_id_equality(self) -> None:
...
+            Key.from_purl(self.sample_purl), Key.from_cpe(self.sample_cpe)
+        )
+        self.assertEqual(purl_only, purl_and_cpe)
+        self.assertEqual(hash(purl_only), hash(purl_and_cpe))
+
+    def test_usable_as_dict_key_with_different_key_sets(self) -> None:
+        # Verifies that dict lookup works correctly when a stored key and the lookup key
+        # compare equal via __eq__ but have different underlying key sets.
+        # (This was a latent bug when the frozen-dataclass-generated __hash__ was used.)
+        id_full = ComponentIdentity(Key.from_purl(self.sample_purl), Key.from_cpe(self.sample_cpe))
```
italvi, line 241:
```suggestion
    def test_usable_as_dict_key_with_different_key_sets(self) -> None:
        id_full = ComponentIdentity(Key.from_purl(self.sample_purl), Key.from_cpe(self.sample_cpe))
```

italvi (approved):

PR #432, opened 2025-05-12
feat: update schema so device do no longer require licenses or copyright

### LINKED ISSUE #405: Custom Schema: Device should not  require licenses
A component of type "device" represents hardware. Hardware is typically owned, not licensed, so the licenses field should not be required.

CBeck-96:
closes #405


italvi (changes requested):
You still require `copyright` in the `metadata.component` when the `type` is device, even though it should not be the case. According to google, you can't claim a copyright on the hardware itself (only some rights on the design).

PR #433, opened 2025-05-12
feat: make licensor and licensee optional for type appliance

### LINKED ISSUE #406: Custom Schema: Appliance should not require licensee

CBeck-96:
closes #406

### DIFF
--- a/tests/test_validate.py
+++ b/tests/test_validate.py
@@ -637,6 +637,26 @@ class TestValidateLicensing(unittest.TestCase):
             issues = validate_test(sbom)
             self.assertEqual(search_for_word_issues("licensee", issues), True)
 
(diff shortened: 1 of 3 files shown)


tests/test_validate.py
```diff
@@ -675,6 +675,26 @@ def test_no_licensee(self) -> None:
             issues = validate_test(sbom)
             self.assertEqual(search_for_word_issues("licensee", issues), True)
 
+    def test_licensetype_appliance_no_licensor_or_licensee(self) -> None:
```
italvi, line 678: @CBeck-96 why only a test for `components` but not `metadata.component`?

italvi (approved):

PR #438, opened 2025-05-20
feat: add email to init sbom arguments and compositions field

CBeck-96:


italvi (changes requested):

cdxev/initialize_sbom.py
```diff
@@ -48,11 +53,20 @@ def initialize_sbom(
...
-    )
+    if email is None:
+        metadata_authors = OrganizationalContact(
+            name=authors,
+        )
+    else:
+        try:
+            validate_email(email, check_deliverability=False)
+        except EmailNotValidError:
+            raise ValueError("Provided email is invalid.")
```
italvi, line 64: you are throwing an whole traceback this way
image

You have to use a `try-except`-block within the main, see e.g. for [set](https://github.com/Festo-se/cyclonedx-editor-validator/blob/eda2333d1a7bad4d7f66dafacefa485172b64b88/cdxev/__main__.py#L939).

You could do something like:
```python
def invoke_init_sbom(args: argparse.Namespace) -> int:
    try:
        sbom = initialize_sbom(
            software_name=args.name,
            authors=args.authors,
            supplier=args.supplier,
            version=args.version,
            email=args.email,
        )
    except ValueError as exc:
        print(f"Error: {exc}")
        return Status.USAGE_ERROR
    write_sbom(sbom, args.output, update_metadata=False)
    return Status.OK
```

italvi (changes requested):
See comment, have to adjust the handling of the ValueError, as you otherwise show the whole traceback.

italvi (approved):
lgtm now, thanks!

PR #473, opened 2025-07-26
chore: move from poetry to uv

italvi:


italvi:
@mmarseu 
Previously
image

Now:
image

mmarseu (dismissed):
I can't tell whether the Github workflows work correctly but I'm assuming you tested them. There are two minor comments, nothing which absolutely requires changes.

Additionally, you might want to consider adding a uv hook to pre-commit, e.g.
```yaml
- repo: https://github.com/astral-sh/uv-pre-commit
    # uv version.
    rev: <uv-version>
    hooks:
      - id: uv-lock
```
The downside being that there is no easy mechanism to keep the used uv version in sync with what devs use. Not that I know of, anyways, but uv moves very quickly.

Also consider adding a `.python-version` file (content is simply a version string `x.y.z`) to ensure uv uses the same python version for all devs.

All of these are only suggestions. You decide what to do with them 😄

mmarseu (approved):

PR #602, opened 2025-12-29
fix: adjust check for festo in  copyright and supplier field

italvi:
Should be merged and reviewed afer #594 was merged.

PR #692, opened 2026-05-02
feat: add support for CycloneDX 1.7 and cryptography definitions schema

italvi:
MR should be only merged and reviewed after #602 was merged.


italvi:
@CBeck-96 right now the test is failing, as I already adjusted the validation of copyright for 1.7. That's why the MR should be only merged and reviewed after #602 was merged.

CBeck-96 (approved):

PR #731, opened 2026-06-01
691 feat support regex for id in set command

### LINKED ISSUE #691: feat: support regex for id in set command
It should be possible to provide a regex in the id (name, purl, cpe) in the set-command. However, this should only be possible when setting a flag and not happen by default.

CBeck-96:
Implements the regex functionality for the set command.
closes #691


docs/source/usage/set.rst
mmarseu, line 1: Firstly, the added information seems to be duplicated in two different places.
Secondly, please integrate the important bits into the *Target components* section. This puts all the relevant info into one place and avoids further duplication (for instance, *Target components* already explains that only a single identifier may be specified for any given update, so there is no need to repeat that for regexes).

I guess, my central critique is that regex matching shouldn't be like a completely different mode which changes everything about how target components are identified (hence my dislike for a `--allow-regex` flag). It is simply an additional way to identify a component all while staying within the same framework already established and known to users.

cdxev/set.py
```diff
@@ -165,6 +167,88 @@ def from_coordinates(
 
 _IDENTIFIERS = {"cpe", "purl", "swid", "name", "version", "group"}
 _PROTECTED = _IDENTIFIERS | {"components"}
+# Ordered list — iteration order must be deterministic.
+_REGEX_IDENTIFIERS = ["cpe", "purl", "name"]
```
mmarseu: Any particular reason why *group* isn't allowed?

mmarseu (changes requested):
I would much prefer an implementation without a flag which enables regex interpretation for all updates specified in a set file. Seems too risky in large files and it might even put users into a position where they cannot use the feature because their set file happens to specify an update to an unfortunately named component which is a valid regex.
It would be more reliable and more consistent with the way version ranges are specified, if users had to explicitly opt into regex evaluation on a per-identifier basis.

An example could be:
```json
[
  {
    "id": {
      "name": "explicitly-named-component"
    },
    "set": {  }
  },
  {
    "id": {
      "namePattern": "regex-pattern.*"
    },
    "set": {  }
  },
  {
    "id": {
      "name": {
        "regex": "regex-pattern.*"
      }
    },
    "set": {  }
  },
]
```

mmarseu (approved):

PR #736, opened 2026-06-06
chore: remove unnecessary Dependabot pre-commit updater workflow and add config for only one PR for pre-commit

italvi:
We added update of pre-commit via dependabot, but did not remove the back then generated file, which now caused duplicate PRs, see #720 and #723 or #739 and #740.

Also, I added the config that PRs for pre-commit are only done within one PR.


mmarseu:
Those two workflows do different things, so I'm not sure they are the reason for duplicate PRs.
The new dependabot feature _updates hooks_. The old workflow doesn't do that. It ensures that the _type stubs_ which are part of our dev dependencies are synchronized to the `additional_dependencies` property of the pre-commit mypy hook.

italvi:
> Those two workflows do different things, so I'm not sure they are the reason for duplicate PRs. The new dependabot feature _updates hooks_. The old workflow doesn't do that. It ensures that the _type stubs_ which are part of our dev dependencies are synchronized to the `additional_dependencies` property of the pre-commit mypy hook.

But you are "just" updating the `additional_dependencies` with your workflow, which is now also done by dependabot itself, no?

mmarseu (approved):
I retract my prior comment. Turns out dependabot *does* do what the old workflow has done before.

PR #748, opened 2026-06-16
feat: merge metadata.tools

### LINKED ISSUE #336: `merge` doesn't merge the tools of its inputs
_Issue brought up in https://github.com/Festo-se/cyclonedx-editor-validator/issues/334#issuecomment-2530723783_

The `merge` command should include the tools from all of its inputs in the final output. Due to the change made to the spec in v1.5 this means that it might need to convert from the [old tools array](https://cyclonedx.org/docs/1.5/json/#tab-pane_metadata_tools_oneOf_i1) to the [new tools object](https://cyclonedx.org/docs/1.5/json/#tab-pane_metadata_tools_oneOf_i0).

CBeck-96:
closes #336


italvi (approved):
There are several things, where I would say it's not really required, as we should assume that we get a valid SBOM rather than us checking if it's really valid by using `isInstance`, but it solves the issue, so LGTM

PR #749, opened 2026-06-19
chore: add fuzzing to pipeline

italvi:


italvi:
@CBeck-96 please review, the changes I made to the "normal" files are from findings of the fuzzers, which detected not proper error handling for wrong regex or wrong version. However, as it took 15min to pass, I removed the PR workflow, and will let it rather run only as nightly scan on main

CBeck-96 (approved):

PR #756, opened 2026-06-22
feat: prepend parent bom-refs in hierarchical merge

### LINKED ISSUE #385: Prepend parent bom-refs in hierarchical merge
Conventionally, in CycloneDX, the bom-ref is used by many tools to denote a sort of path to a component in the component hierarchy.
E. g., 
```json
{
  "bom-ref": "compA",
  "name": "compA",
  "components": [
    {
      "bom-ref": "compA/subcompA",
      "name": "subcompA"
    }
  ]
}
```

It's a useful convention, because it allows users to easily identify the location of a component in the product.

Hierarchical merges mess with this convention because they move components to a deeper level in the resulting SBOM. If the bom-refs in the original SBOMs were path-like (such as the example above), in the result, they would no longer be correct.
This problem could be solved by prepending the bom-ref of the new parent when doing a hierarchical merge.

CBeck-96:
closes #385


italvi:
@CBeck-96 you linked #758 to this MR, but you are not using/extending amend?

CBeck-96:
> @CBeck-96 you linked #758 to this MR, but you are not using/extending amend?

corrected

italvi (changes requested):
Can you please adjust the [documentation](https://github.com/Festo-se/cyclonedx-editor-validator/blob/385-prepend-parent-bom-refs-in-hierarchical-merge/docs/source/usage/merge.rst) accordingly, to also describe the implemented behavior there? But apart from them, nothing to add, and I checked the tests, and lgtm and seems to resolve the issue as described in the issue.

italvi:
@mmarseu could you please confirm that the current implementation was done as desired? I.e. with only using "/" as indicator for separation.

mmarseu:
I've updated a test because IMO it was only half as useful as it could have been. Don't worry, it still passes ;)
I've added another, failing test. This isn't actually related to this PR just something I've discovered while trying it out. Feel free to cherry-pick it to another branch if you don't want to fix it here.

Other than that, why did you chose to limit this algorithm to bom-refs which are already path-like? Any reason why we shouldn't do it for any other? That would significantly simplify the logic for no apparent downside.

italvi (changes requested):
As suggest by @mmarseu 
> Other than that, why did you chose to limit this algorithm to bom-refs which are already path-like? Any reason why we shouldn't do it for any other? That would significantly simplify the logic for no apparent downside.

Please adjust the logic accordingly.

mmarseu (changes requested):
Thank you for the previous changes to this PR.
There are more issues, though, some of which are once again not directly related to this PR but to the merge feature in general. Essentially, the problem is that the code only deduplicates a very few select instances of bom-refs but the CycloneDX 1.7 schema allows bom-refs in a total of 34 different places.
I've added two (almost identical) failing tests. They fail, if any merged fields contain duplicate bom-refs.
It would likely be more efficient to scan the entire dict for fields named "bom-ref" than to craft specialized code for all element types which may contain bom-refs.

On a second note, your merge module previously already contained functions for bom-ref deduplication (`make_bom_refs_unique()`) and this PR adds another function (`_ensure_unique_bom_ref()`) for more or less the same purpose. Can you explain why that is necessary?

mmarseu (approved):
Due to rebases I have no clue what has changed since my last review. Sorry that I cannot review the entire PR again. I will simply assume that you implemented the changes I asked for last time and sign off on this.

italvi (approved):

PR #759, opened 2026-06-23
fix: adapt vulnerability identities when merging several SBOMs

CBeck-96:
When merging several SBOMs the references used in affects of a vulnerability can be overwrittten. Since the identities are determined in the beginning, this can lead to a crash.


italvi (approved):
tested it with the internal vex we received and the previous issue was resolved, so lgtm

Issue #767, opened 2026-07-03
Validation of file names could be more specific


CBeck-96:
The validation using the provided custom regex could specify exactly where an error occurs and what names would be legitimate

italvi:
Due to changes done in #808 this is not valid anymore, as we removed the regex validation, now only somebody could provide a regex on their own and they should know then if it's valid.

PR #768, opened 2026-07-03
feat: improve custom filename error messages

CBeck-96:


italvi:
I will close this one, as we also closed the corresponding issue #767, as we removed the custom filename

PR #789, opened 2026-07-17
fix: correct treatment of swid objects across the tool

CBeck-96:
The comparison if SWID object crashes the application.


tests/test_identity.py
```diff
@@ -108,6 +108,26 @@ def test_swid_creation(self) -> None:
...
+    def test_swid_inequality_when_version_present_on_one_side(self) -> None:
+        left = Key.from_swid({"tagId": "alpine", "name": "alpine"})
+        right = Key.from_swid({"tagId": "alpine", "name": "alpine", "version": "3.23.3"})
+        self.assertNotEqual(left, right)
+
+    def test_swid_inequality_without_tagid(self) -> None:
+        left = Key.from_swid({"name": "alpine"})
+        right = Key.from_swid({"name": "alpine"})
+        self.assertNotEqual(left, right)
+
```
italvi, line 136: Can you please use the `sample_swid` as done in the other [tests](https://github.com/Festo-se/cyclonedx-editor-validator/blob/ab8b68d60fc988680ed3eacf745c8bb8857eb91b/tests/test_identity.py#L103)?

cdxev/auxiliary/sbom_functions.py
```diff
@@ -138,6 +138,18 @@ def compare_components(first_component: dict, second_component: dict) -> bool:
...
+def _swid_equals(first_swid: Any, second_swid: Any) -> bool:
+    """Compare SWID values using identity semantics with a permissive fallback.
+
+    SWID comparison in the identity layer supports mapping and JSON-string forms. For malformed
+    or non-standard values, preserve historic behavior via case-insensitive string comparison.
+    """
+    try:
+        return Key.from_swid(first_swid) == Key.from_swid(second_swid)
+    except (TypeError, ValueError, KeyError, json.JSONDecodeError):
+        return str(first_swid).lower() == str(second_swid).lower()
```
italvi: As you are using `swid_equals(first_component.get("swid")` you will always get either a (hopefully valid, as this is something coming mainly from package managers) swid-object or `None`, so this does not make sense. You also don't have this case/branch in your added tests.

italvi (changes requested):

cdxev/auxiliary/sbom_functions.py
```diff
@@ -117,7 +117,9 @@ def compare_components(first_component: dict, second_component: dict) -> bool:
         else:
             return False
     if first_component.get("swid", "") and second_component.get("swid", ""):
-        if first_component.get("swid", "1").lower() == second_component.get("swid", "2").lower():
+        if Key.from_swid(first_component["swid"]) == Key.from_swid(
+            second_component["swid"]
+        ):
             is_equal = True
         else:
             return False
```
italvi: Why not just `return Key.from_swid(first_component["swid"]) == Key.from_swid(second_component["swid"])`? Has the same effect as having it in the if/else (you could do that for all the if/else, instead of doing == and then return true, as if equal it will also be of value true)

italvi (changes requested):

cdxev/auxiliary/sbom_functions.py
```diff
@@ -107,15 +107,17 @@ def compare_components(first_component: dict, second_component: dict) -> bool:
     """
     is_equal = False
     if first_component.get("purl", "") and second_component.get("purl", ""):
-        return first_component.get("purl", "1").lower() == second_component.get("purl", "2").lower()
+        return bool(
+            first_component.get("purl", "1").lower() == second_component.get("purl", "2").lower()
+        )
```
italvi, line 112: why the additional conversion with bool()?

cdxev/auxiliary/sbom_functions.py
```diff
@@ -107,15 +107,17 @@ def compare_components(first_component: dict, second_component: dict) -> bool:
...
 
     if first_component.get("cpe", "") and second_component.get("cpe", ""):
-        return first_component.get("cpe", "1").lower() == second_component.get("cpe", "2").lower()
+        return bool(
+            first_component.get("cpe", "1").lower() == second_component.get("cpe", "2").lower()
+        )
```
italvi, line 117: why the additional conversion with bool()?

cdxev/auxiliary/sbom_functions.py
CBeck-96: Currently it will return false if any of purl, cpe, swid do not align
with this change, it will enforce a hierarchy
same purl overules different cpe

cdxev/auxiliary/sbom_functions.py
CBeck-96: mypy

cdxev/auxiliary/sbom_functions.py
CBeck-96: mypy

tests/test_sbom_functions.py
italvi, line 1: Image
swid requires tagId and name, can you please add `name`, so that we are compliant to the spec? You have it in this file several times, and did not want to comment on every single occasion.

italvi (changes requested):

italvi (approved):

PR #808, opened 2026-07-29
feat: update filename validation to allow only specific patterns and adhere to cdx pattern

### LINKED ISSUE #807: feat: remove strict restrictions for filename validation

italvi:


cdxev/validator/helper.py
```diff
@@ -109,10 +108,7 @@ def validate_filename(
     schema_type: t.Optional[str],
 ) -> t.Union[t.Literal[False], str]:
     if not regex:
-        if schema_type == "custom":
```
CBeck-96, line 112: Function still takes the arguments sbom and schema_type, now unused

CBeck-96 (changes requested):

cdxev/validator/helper.py
italvi: should be fixed now through https://github.com/Festo-se/cyclonedx-editor-validator/pull/808/changes/8996f24ae1aa8309c17cac92f16931a34d62226d, also adjusted the tests there accordingly, as less parameters required

CBeck-96 (approved):

PR #809, opened 2026-07-31
fix: add patent  assertions to custom 1.7 schema

CBeck-96:
Custom Schem throws errors when patentAssertions are present.


italvi (changes requested):
Hi @CBeck-96, when checking your change, I observed that you even missed one place where "patentassertions" should be there, according to the vanilla schema. Also I was a little worried, that we might have omitted other fields too - which we did. I did now, with the help of copilot, a check of all fields, and yeah, there were more. Like tlpclassification, accepted hashes were missing, a lot of tls/crypto stuff was not there. The new custom schema file is here: [bom-1.7-custom.schema.json](https://github.com/user-attachments/files/31302937/bom-1.7-custom.schema.json). I didn't want to push it, as otherwise we won't be able to merge this PR, without the approval of a third person. Feel free to do a diff and confirm the changes for yourself.

italvi (approved):

PR #817, opened 2026-08-11
fix: add missing spaces in help text

Gronner:
A small change adding spaces to help texts, so they don't appear concatenated.


italvi (approved):
Hello @Gronner, thanks for your first contribution and nice catch!

PR #826, opened 2026-08-21
refactor: migrate to jsonschema-rs

### LINKED ISSUE #695: feat: Migrate to jsonschema-rs for efficiency?
I've been pinged by a coworker today about validation failing for a very large SBOM.

I've reproduced this on my machine with their SBOM which weighs in at 25 MB, with about 58k components. I can't share it publicly but you can drop me a message if you want it.

Some testing indicates that the issue doesn't stem from our post-processing logic. Even using the semi-official `check-jsonschema` CLI version of `python-jsonschema`, validation has been running for like 20 minutes on my PC with no end in sight.

Using [jsonschema-rs](https://pypi.org/project/jsonschema-rs/) as an alternative implementation, I've been able to get results for the same SBOM. It's not fast by any measure but its usable.

Validation against the original CycloneDX schema is almost instantaneous and produces 2 errors.
Validation against the custom schema takes about 30 seconds and produced a whopping 175k errors.

I wasn't able to determine whether the limiting factor is schema complexity (which is much higher in our custom schema than in the original) or error count. But in any case, the `pythonschema` package clearly fails to validate this SBOM.

Question is, do we want to support validation of SBOM with such outrageous numbers of errors? If so, we might need to move to an alternative implementation (such as `jsonschema-rs`). 
Alternatively, we might monitor the validation and abort it if it doesn't finish, telling the user that their SBOM has too many issues to be validated.

italvi:
~Should be reviewed and merged after #825 was merged.~


italvi:
@mmarseu can you please review and verify that your test done in #695 is still valid and the results are faster for larger SBOMS than for the previous implementation with `jsonschema`?

mmarseu (approved):
lgtm

PR #832, opened 2026-08-25
feat: support component manufacturer field as equivalent to supplier

### LINKED ISSUE #814: Support component manufacturer field
The `authors` and `manufacturer` properties have been introduced in CycloneDX 1.6 to replace `author` and supplement the existing `author`, `supplier`, and `publisher` properties, respectively.

Should we treat them as equivalent to the existing fields and add support for them to the `validate` and `amend` commands?

italvi:
~Should be reviewed and merged after #826 was merged.~


cdxev/amend/operations.py
```diff
@@ -276,6 +278,30 @@ def infer_supplier(self, component: dict) -> None:
 
         if "publisher" in component:
             supplier["name"] = component["publisher"]
+        elif "manufacturer" in component and isinstance(component["manufacturer"], dict):
+            manufacturer = component["manufacturer"]
+            if "name" in manufacturer:
```
CBeck-96: What about contact?
It is anyOf name, url, contact

If url and name are empty, supplier will be empty since authors is skipped due to elif structure. The contact data should also be extracted

CBeck-96 (changes requested):

CBeck-96 (approved):

PR #833, opened 2026-08-27
feat: validate license expressions

### LINKED ISSUE #554: Validate license expressions
Licenses in CycloneDX can be formulated with SPDX expressions.

Example:
```json
{
  "name": "foo",
  "version": "1.0",
  "licenses": [
    {
      "expression": "GPL-2.0 or MIT"
    }
  ]
}
```

Because there is no JSON schema definition of SPDX expressions, really, one can put anything in there and the validator would be fine with it.
I've seen cases where this loophole was used to remove validation errors.

How about we use the [license_expression](https://pypi.org/project/license-expression/) package to validate such expressions in addition to the JSON schema validation?

italvi:
~Should be reviewed and merged after #832 was merged.~


tests/test_validate.py
```diff
@@ -436,13 +436,41 @@ def test_components_component_license_or_expression(self) -> None:
...
             sbom["specVersion"] = spec_version
-            sbom["components"][0]["licenses"] = [{"expression": "Apache-2.0"}]
+            sbom["components"][0]["licenses"] = [{"expression": "LicenseRef-custom"}]
             issues = validate_test(sbom)
             self.assertEqual(issues, ["no issue"])
 
+    def test_components_invalid_license_expression(self) -> None:
+        expressions = [
+            "GPL-2.0 or",
+            "not-a-license",
```
CBeck-96, line 464: The default schema has no minimal length (custom has), so with the default schema an empty string "" passes. Should this be caught as well?

cdxev/validator/helper.py
```diff
@@ -195,6 +199,42 @@ def sort_key(issue: ValidationIssue) -> tuple[str, str, str]:
     )
 
 
+def _validate_license_expressions(instance: dict[str, t.Any]) -> list[ValidationIssue]:
+    issues: list[ValidationIssue] = []
+    for path in _license_expression_paths(instance):
+        expression = t.cast(str, _value_at_path(instance, path))
+        try:
+            parsed = _LICENSE_EXPRESSION_PARSER.parse(expression)
```
CBeck-96, line 208: Several licenses that are accepted as valid ids in spdx.schema.json e.g. "sqlitestudio-OpenSSL-exception" are not accepted as expressions by the parser. Add comparison with the list for unkonwn keys before raising the error?

cdxev/validator/helper.py
```diff
@@ -195,6 +199,42 @@ def sort_key(issue: ValidationIssue) -> tuple[str, str, str]:
...
+            unknown_keys = _LICENSE_EXPRESSION_PARSER.unknown_license_keys(parsed)
+            invalid_keys = [key for key in unknown_keys if not _LICENSE_REF.fullmatch(key)]
```
CBeck-96:
```suggestion
            invalid_keys = [
                key
                for key in unknown_keys
                if key not in _SPDX_LICENSE_IDS and not _LICENSE_REF.fullmatch(key)
            ]
```

cdxev/validator/helper.py
```diff
@@ -139,6 +141,7 @@ def validate_filename(
     "vulnerabilities": "vulnerability",
 }
 _ENTITY_COLLECTIONS = {"components", "dependencies", "services", "vulnerabilities"}
+_LICENSE_EXPRESSION_PARSER = get_spdx_licensing()
```
CBeck-96:
```suggestion
_SPDX_LICENSE_IDS = frozenset(t.cast(list[str], load_spdx_schema()["enum"]))
```

tests/test_validate.py
```diff
@@ -436,13 +436,41 @@ def test_components_component_license_or_expression(self) -> None:
...
+        ]
+        for expression in expressions:
+            for spec_version in list_of_spec_versions:
+                sbom = get_test_sbom()
+                sbom["specVersion"] = spec_version
+                sbom["components"][0]["licenses"] = [{"expression": expression}]
+                issues = validate_test(sbom)
+                self.assertTrue(search_for_word_issues("not a valid license expression", issues))
+
```
CBeck-96, line 473:
```suggestion
    def test_components_license_ids_are_valid_expressions(self) -> None:
        with open(
            Path(__file__).parent.parent / "cdxev" / "auxiliary" / "schema" / "spdx.schema.json",
            encoding="utf-8",
        ) as schema_file:
            spdx_schema = json.load(schema_file)

        for license_id in spdx_schema["enum"]:
            for spec_version in list_of_spec_versions:
                with self.subTest(license_id=license_id, spec_version=spec_version):
                    sbom = get_test_sbom()
                    sbom["specVersion"] = spec_version
                    sbom["components"][0]["licenses"] = [{"expression": license_id}]
                    issues = validate_test(sbom)
                    self.assertEqual(issues, ["no issue"])
```

CBeck-96 (changes requested):

tests/test_validate.py
italvi: As the default schema allows it, we should not make it strict by implementing own rules. Otherwise we must do also things like making version mandatory, which then would make the provision of a custom schema somehow abundant.

cdxev/validator/helper.py
italvi: As this is now handled by introducing the new spdx variable, I think this one can be resolved.

tests/test_validate.py
italvi: We are already doing a test like this [here](https://github.com/Festo-se/cyclonedx-editor-validator/pull/833/changes#diff-b77d003dbd8ec8f0b12435dfb6950b341c9dc78cc083cf851c5259142286a210R440) with just checking for one ID, so it feels redundant to do it for every single spdx id.

italvi:
@CBeck-96 please check my comments on the unresolved threads, do a review again, and resolve the threads if you agree.  Otherwise feel free to discuss.

CBeck-96 (approved):

PR #834, opened 2026-08-28
feat: prepend parent bom refs in nested components

### LINKED ISSUE #758: Prepend parent bom-refs in hierarchical components with amend
Extend amend to provide a operation that goes through the SBOM and prepends parent components with "/".

italvi:
Should be reviewed and merged after #833 was merged.


cdxev/auxiliary/sbom_functions.py
```diff
@@ -198,6 +200,160 @@ def extract_components(list_of_components: Sequence[dict]) -> Sequence[dict]:
...
+            _collect_schema_reference_fields(
+                nested_value, schema_name, schema_dir, schemas, fields
+            )
+    elif isinstance(value, list):
+        for item in value:
+            _collect_schema_reference_fields(item, schema_name, schema_dir, schemas, fields)
+
+
+@lru_cache(maxsize=None)
+def _cached_schema_reference_fields(spec_version: Optional[str]) -> tuple[str, ...]:
```
CBeck-96: The name is generic but the function hardcods fields: "bom-ref". This should be a function parameter or shared object etc.

tests/test_amend.py
```diff
@@ -579,5 +581,222 @@ def test_dont_delete_name_with_text(self):
         self.assertDictEqual(self.component, expected)
 
 
+class HierarchicalBomRefsTestCase(unittest.TestCase):
+    def test_supports_every_cyclonedx_version(self) -> None:
+        for spec_version in ("1.2", "1.3", "1.4", "1.5", "1.6", "1.7"):
```
CBeck-96: Would make this a global variable, since it must be updated on the regular.
Since other tests also require the list, it might make sense to create one in the helper file.

tests/test_amend.py
```diff
@@ -579,5 +581,222 @@ def test_dont_delete_name_with_text(self):
...
+                    "name": "root",
+                    "bom-ref": "module-a",
+                    "components": [
+                        {
+                            "name": "first child",
+                            "bom-ref": "library-b",
+                            "components": [
+                                {
+                                    "name": "second child",
+                                    "bom-ref": "application-c",
```
CBeck-96, line 636: Why not numbers als refs like above/ why above numbers instead of those?

italvi: chosen names instead of numbers

tests/test_amend.py
italvi: done, just like it's done in `test_validate`

cdxev/auxiliary/sbom_functions.py
italvi: changed to field being a variable, so that also in the feature we might adjust it

CBeck-96 (approved):

PR #854, opened 2026-09-14
fix: loosen copyright pattern validation to reflect intentional behavior

italvi:


mmarseu (approved):
Approved, provided you're okay with the tests I've added. Please look them over and if you feel that every case on the accepted and rejected lists should really be accepted or rejected, respectively, then go ahead and merge.

PR #855, opened 2026-09-18
refactor: use operation structure for hierarchical bom ref

italvi:


mmarseu (changes requested):
I appreciate the effort but this PR doesn't entirely solve the underlying problem - if you want to call it a problem.

It shifts parts of the logic around so that the operation more closely adheres to the design paradigm of operation classes and this does solve an issue or two. 
But it still loops through all components in its prepare step which is not ideal for two reasons:
1. Performance suffers if a bunch of operations each loop over thousands of components.
2. Subtle bugs might be introduced by each operation implementing its looping logic a little differently.

Both of those issues are minor and you might consider them to be acceptable but I have a counter-proposition which I still consider superior 😉 

I haven't thought it through entirely, but I believe it would be rather straight-forward to eliminate the up-front loop in the prepare step. We know that the algorithm travers the tree top-down and we know that the metadata component comes before every other component. So on processing any component, we can store references to its children as we go along. Once its the child's turn to be passed to `handle_component`, we're sure, the parent has already been processed and the link between the child and its parent has been stored.
Sure, this still means we need to loop over child components in `handle_component` but at least we don't descend down the tree.

mmarseu:
On a related note, this entire PR was triggered by the question of interactions between this operation and `AddBomRef`. 
~With this new design, those interactions exist, where in the old one, they didn't.~ Edit: I meant in the design I propose above, they exist. In the original design or in the design as implemented in this PR, I believe they don't. 

An easy and IMO acceptable solution might be to add some documentation to both operations. If both are enabled on the same run, the order in which they appear on the command line governs which one runs first. In most cases, `AddBomRef` should probably run first to ensure consistent results.

mmarseu:
And finally, why hard-coding the path separator used in the bomrefs? Operations can accept arguments. Why not leave it up the user?

cdxev/auxiliary/sbom_functions.py
```diff
@@ -434,9 +436,12 @@ def _recurse(
...
                 _recurse(component["components"], func, *args)
 
-    if not skip_meta:
-        if component := sbom.get("metadata", {}).get("component", None):
-            func(component, *args, **kwargs)
+    metadata_component = sbom.get("metadata", {}).get("component")
+    if metadata_component:
+        if not skip_meta:
+            func(metadata_component, *args, **kwargs)
+        elif "components" in metadata_component:
```
mmarseu: Why apply the function to subcomponents of the metadata component? First of all, this field should likely never be used (subcomponents of the metadata component are the components of the SBOM, after all).
Secondly, when the argument `skip_meta` is given, the user wouldn't expect the function to skip the metadata component but still apply to its children, right?

cdxev/amend/operations.py
```diff
@@ -523,7 +525,8 @@ class HierarchicalBomRefs(Operation):
     """
     Prepends parent bom-refs to the bom-refs of nested components.
```
mmarseu, line 526: Add a paragraph about the interactions between this and `AddBomRef` to the class docstring.

mmarseu (changes requested):
Almost good to go. I've still got two minor issues :)

mmarseu (approved):
lgtm

